PR #28, opened 2020-03-25
taking the first message if fits

Eransho:


src/main/java/com/tabnine/TabNineCompletionContributor.java
```diff
@@ -198,7 +203,10 @@ public void fillCompletionVariants(CompletionParameters parameters, CompletionRe
             result = result.withRelevanceSorter(CompletionSorter.defaultSorter(parameters, originalMatcher).weigh(new TabNineWeigher()));
             result.restartCompletionOnAnyPrefixChange();
             if (completions.user_message.length >= 1) {
-                result.addLookupAdvertisement(String.join("\n", completions.user_message));
+                String firstMsg = completions.user_message[0];
+                if (firstMsg.length() <= ADVERTISEMENT_MAX_LENGTH) {
```
amircodota: Don't we want to cut it if it is too long?

Eransho: I don't see the benefit of display incomplete sentence. This is not a perfect solution but I think that a perfect solution doesn't worth the cost for such minor feature.
Other acceptable alternative will be to not use the advertisement at all

amircodota: OK, makes sense

PR #33, opened 2020-05-10
Adding plugin metadata

Eransho:


Eransho:
>     1. Need to change this to the new Vec based command line approach.
> 
>     2. Need to test what happens to an existing client that will upgrade to this version.
>        My concern is that it will not ownload a new binary and fail to run the old one

Didn't understand 1. As for 2, how can we test it?

amircodota:
@Eransho 
1. You changed how the command line arguments are passed, right?
2. I think install from master, and have the plugin downloaded. Then run from your branch and see what happens

Eransho:
@amircodota 

1. These changes are aligned to the changes in the binary
2. The newer plugin version (this one) will not work with older binary because of the new command argument - is that what you meant?

amircodota:
1. Ah OK. I thought you need to merge each of the metadata arguments with --client-metadata, not just the first one
2. Yes, but it will not download a newer version either, leaving the user with a non-functioning setup

PR #34, opened 2020-07-15
Adding plugin state listener

Eransho:
Handling plugin uninstall - https://codota.atlassian.net/browse/DEV-384?atlOrigin=eyJpIjoiOTNlYjc1ZWRjM2MyNGVhZTg5MTVmOWZiNWJhZjUxY2UiLCJwIjoiaiJ9

### DIFF
--- a/src/main/java/com/tabnine/lifecycle/TabNinePluginStateListener.java
+++ b/src/main/java/com/tabnine/lifecycle/TabNinePluginStateListener.java
@@ -1,0 +1,31 @@
+package com.tabnine.lifecycle;
+
+import com.intellij.ide.plugins.IdeaPluginDescriptor;
+import com.intellij.ide.plugins.PluginStateListener;
+import com.intellij.openapi.diagnostic.Logger;
+import com.intellij.openapi.extensions.PluginId;
+import com.tabnine.TabNineProcess;
+import com.tabnine.Utils;
+import org.jetbrains.annotations.NotNull;
+
+import java.io.IOException;
+import java.util.Collections;
+
+public class TabNinePluginStateListener implements PluginStateListener {
+
+    @Override
+    public void install(@NotNull IdeaPluginDescriptor descriptor) {
+    }
+
+    @Override
+    public void uninstall(@NotNull IdeaPluginDescriptor descriptor) {
+        final PluginId pluginId = descriptor.getPluginId();
+        if (pluginId != null && pluginId.equals(Utils.getPluginId())) {
+            try {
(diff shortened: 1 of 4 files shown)


src/main/java/com/tabnine/lifecycle/TabNinePluginStateListener.java
```diff
@@ -0,0 +1,31 @@
...
+    @Override
+    public void install(@NotNull IdeaPluginDescriptor descriptor) {
+    }
+
+    @Override
+    public void uninstall(@NotNull IdeaPluginDescriptor descriptor) {
+        final PluginId pluginId = descriptor.getPluginId();
+        if (pluginId != null && pluginId.equals(Utils.getPluginId())) {
+            try {
+                TabNineProcess.startTabNine(true, Collections.singletonList("--uninstalling"));
```
amircodota, line 25: just make sure we need --uninstalling and not --uninstalled

Eransho: Yes, that's what we use in sublime plugin

src/main/java/com/tabnine/lifecycle/TabNinePluginStateListener.java
```diff
@@ -0,0 +1,31 @@
...
+import java.util.Collections;
+
+public class TabNinePluginStateListener implements PluginStateListener {
+
+    @Override
+    public void install(@NotNull IdeaPluginDescriptor descriptor) {
+    }
+
+    @Override
+    public void uninstall(@NotNull IdeaPluginDescriptor descriptor) {
```
shaiahr, line 21: What about disable?

Eransho:
> I think T9 supports 'hot deploy' (i.e. it can be enabled without restart). Will the listener be registered in that case?

How is that related to the uninstall event?

src/main/java/com/tabnine/lifecycle/TabNinePluginStateListener.java
Eransho: Disable will be handled separately. It requires changes in the binary as well

shaiahr:
> > I think T9 supports 'hot deploy' (i.e. it can be enabled without restart). Will the listener be registered in that case?
> 
> How is that related to the uninstall event?

In Codota, we report both. Don't we want to do the same here?

boaz-codota:
https://github.com/codota/tabnine-intellij/pull/54

PR #39, opened 2020-09-08
Add no-lsp flag

boaz-codota:


src/main/java/com/tabnine/TabNineProcess.java
```diff
@@ -33,7 +32,7 @@ void startTabNine() throws IOException {
         // setup since we assume it's already built into the IDE
         List<String> command = new ArrayList<>();
         command.add(TabNineFinder.getTabNinePath());
-        List<String> metadata = new ArrayList<>();
+        List<String> metadata = new ArrayList<>(asList("--no-lsp", "true"));
```
amircodota, line 35: This is silly, but no_lsp is not metadata. A metadata is a generic key=value parameter, that's passed as is to analytics.

It is better to add it to the command object, around lines 42

amircodota (approved):
Had a minor comment, but other than that LGTM

src/main/java/com/tabnine/TabNineProcess.java
boaz-codota: OK. The separation do looks silly.

boaz-codota:
DO NOT MERGE BEFORE THE TABNINE BINARY IS IN CIRCLE FOR ALMOST EVERYONE (as this breaks the binary).
@amircodota is there even a way to control this?

amircodota:
> DO NOT MERGE BEFORE THE TABNINE BINARY IS IN CIRCLE FOR ALMOST EVERYONE (as this breaks the binary).
> @amircodota is there even a way to control this?

@boaz-codota Not really. Experience has shown that most users upgrade pretty quickly, but there's a long tail of users that can take a long time.

What you can do is check the binary version and only add the flag after a certain version.
Not pretty, but it will work

boaz-codota:
> > DO NOT MERGE BEFORE THE TABNINE BINARY IS IN CIRCLE FOR ALMOST EVERYONE (as this breaks the binary).
> > @amircodota is there even a way to control this?
> 
> @boaz-codota Not really. Experience has shown that most users upgrade pretty quickly, but there's a long tail of users that can take a long time.
> 
> What you can do is check the binary version and only add the flag after a certain version.
> Not pretty, but it will work

How can I do that?

amircodota:
@boaz-codota I think intellij downloads the binary into a directory of its version something like ~/.tabnine/{version}/{platform}/TabNine. You you can check the path.

Another option - run `TabNine --version`

PR #40, opened 2020-09-08
DEV-629 - when no characters were entered but the user summons the su…

avnerbarr:
…ggestions box the suggestions would return empty. Instead this code should return the default suggestions from the IDE

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -4,7 +4,7 @@ plugins {
 }
 
 group 'com.tabnine'
-version '0.1.12'
+version '0.1.12.2'
 
 sourceCompatibility = 1.8
 
@@ -17,9 +17,17 @@ dependencies {
 }
 
 intellij {
-    version '2019.3'
+    // testing intellij-2020.2
+    //    version '2020.2'
+    // testing intellij-2019.1
+    version '2019.1'
+
     type 'IC'
     updateSinceUntilBuild false
+    // testing intellij-2020.2
+    // plugins 'org.rust.lang:0.3.130.3337-202', 'org.toml.lang:0.2.130.3337-202'
+    // testing intellij-2019.1
+    plugins 'org.rust.lang:0.2.98.2125-191'
 }
 
 publishPlugin {
--- a/build.gradle
+++ b/build.gradle
@@ -25,3 +25,9 @@ intellij {
 publishPlugin {
     token project.hasProperty('hubToken') ? hubToken : ''
 }
+
+tasks.create("currentVersion") {
+    doLast {
+        println version
+    }
+}
--- a/src/main/java/com/tabnine/TabNineCompletionContributor.java
+++ b/src/main/java/com/tabnine/TabNineCompletionContributor.java
@@ -199,6 +199,10 @@ public class TabNineCompletionContributor extends CompletionContributor {
         TabNineProcess.AutocompleteResponse completions = this.retrieveCompletions(parameters, baseMaxCompletions);
         PrefixMatcher originalMatcher = result.getPrefixMatcher();
         if (completions != null) {
+            if (originalMatcher.getPrefix().length() == 0 && completions.results.length == 0) {
+//                result.stopHere();
+                return;
+            }
             result = result.withPrefixMatcher(new TabNinePrefixMatcher(originalMatcher.cloneWithPrefix(completions.old_prefix)));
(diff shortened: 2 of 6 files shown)


avnerbarr:
@amircodota I left the verbose code comments purposely for review. will remove them once approved as they are not relevant for public repo

avnerbarr:
option 2 - explicitly call result.runRemainingContributors(parameters, true); and then result.stopHere();

src/main/java/com/tabnine/TabNineCompletionContributor.java
```diff
@@ -210,7 +210,6 @@ public void fillCompletionVariants(CompletionParameters parameters, CompletionRe
                 result.addLookupAdvertisement(details);
             }
             if (originalMatcher.getPrefix().length() == 0 && completions.results.length == 0) {
-                result.stopHere();
                 return;
```
avnerbarr, line 213: option 2 - explicitly call result.runRemainingContributors(parameters, true); and then result.stopHere();

amircodota (approved):
LGTM

src/main/java/com/tabnine/TabNineCompletionContributor.java
```diff
@@ -233,8 +237,9 @@ public void fillCompletionVariants(CompletionParameters parameters, CompletionRe
         }
     }
 
-    @Override
-    public boolean invokeAutoPopup(@NotNull PsiElement position, char typeChar) {
-        return true;
-    }
+//    @Override
```
amircodota, line 240: @avnerbarr I think we need it, but not after every char. We should probably include a-z, digits, <space>, =

Thoughts?

avnerbarr: I played with it and the default behaviour was fine on my system (2020.2.1 CE). 

The method is deprecated and they suggest not using it.

```/**
   * @deprecated Don't use this method, because {@code position} can come from uncommitted PSI and be totally unrelated to the code being currently in the document/editor.
   * Please consider using {@link com.intellij.codeInsight.editorActions.TypedHandlerDelegate#checkAutoPopup} instead.
   */
  @Deprecated
  public boolean invokeAutoPopup(@NotNull PsiElement position, char typeChar) {
    return false;
  }
```

amircodota: OK, cool. Playing with it :+1:

build.gradle
```diff
@@ -17,9 +17,10 @@ dependencies {
 }
 
 intellij {
-    version '2019.3'
```
avnerbarr, line 20: @amircodota Should this be committed?

build.gradle
```diff
@@ -4,7 +4,7 @@ plugins {
 }
 
 group 'com.tabnine'
-version '0.1.12'
+version '0.1.12.2'
```
amircodota, line 7: this should not be merged

build.gradle
amircodota: no

build.gradle
avnerbarr: does the CI/CD upgrade it?

src/main/java/com/tabnine/TabNineCompletionContributor.java
```diff
@@ -209,10 +213,10 @@ public void fillCompletionVariants(CompletionParameters parameters, CompletionRe
                 }
                 result.addLookupAdvertisement(details);
             }
-            if (originalMatcher.getPrefix().length() == 0 && completions.results.length == 0) {
-                result.stopHere();
-                return;
-            }
+//            if (originalMatcher.getPrefix().length() == 0 && completions.results.length == 0) {
```
boaz-codota, line 216: Remove commented code if not neccessary

src/main/java/com/tabnine/TabNineCompletionContributor.java
boaz-codota: Are we sure about this?

avnerbarr:
for some reason git didn't squash my commit when rebased to master 😠

avnerbarr:
@boaz-codota @amircodota closing this PR because of git merge could not squash commits from master (after rebased)

this PR replaces it
https://github.com/codota/tabnine-intellij/pull/44

PR #41, opened 2020-09-13
Add timeout to suggestions requests

boaz-codota:


boaz-codota:
@dimacodota I would like to add tests before merging this, I've pinged eransho for help, he is yet to reply.

src/main/java/com/tabnine/TabNineCompletionContributor.java
```diff
@@ -172,7 +174,7 @@ TabNineProcess getProcOrPrintError() {
                 req.max_num_results = max_num_results;
                 req.region_includes_beginning = (begin == 0);
                 req.region_includes_end = (end == doc.getTextLength());
-                return proc.request(req);
+                return AppExecutorUtil.getAppExecutorService().submit(() -> proc.request(req)).get(1, TimeUnit.SECONDS);
```
dimacodota, line 177: Future. get is throwing a TimeoutException if timed out,  is this handled somewhere ?

dimacodota (commented):
Future. get is throwing a TimeoutException if timed out, is this handled somewhere?

dimacodota:
> @dimacodota I would like to add tests before merging this, I've pinged eransho for help, he is yet to reply.

ping amirtu

src/main/java/com/tabnine/TabNineCompletionContributor.java
boaz-codota: Fuck me. That is why I want tests...

dimacodota: yep, although it is not an easy thing to test...

src/main/java/com/tabnine/TabNineCompletionContributor.java
```diff
@@ -172,7 +174,12 @@ TabNineProcess getProcOrPrintError() {
                 req.max_num_results = max_num_results;
                 req.region_includes_beginning = (begin == 0);
                 req.region_includes_end = (end == doc.getTextLength());
-                return proc.request(req);
+                
+                try {
+                    return AppExecutorUtil.getAppExecutorService().submit(() -> proc.request(req)).get(1, TimeUnit.SECONDS);
```
Eransho, line 179: I think `Application.executeOnPooledThread` is safer (to the best of my knowledge, this is how JB suggesting to run task on pooled thread)

src/main/java/com/tabnine/TabNineCompletionContributor.java
```diff
@@ -172,7 +174,12 @@ TabNineProcess getProcOrPrintError() {
...
+                } catch (Exception e) {
+                    return null;
```
Eransho, line 181: Don't we want to handle exceptions here? 
If not then you might as well catch `Throwable` to cover yourself

Eransho (commented):
See my comments

src/main/java/com/tabnine/TabNineCompletionContributor.java
boaz-codota: It is possible that we would like to add an event here, but I'm not adding events unless explicitly requested anymore (due to latest event bombarding mixpanel). @dimacodota what do you say?

src/main/java/com/tabnine/TabNineCompletionContributor.java
boaz-codota: image
Application is an interface. Did you mean something else? @Eransho

boaz-codota: I found it and changed it.

src/main/java/com/tabnine/TabNineCompletionContributor.java
dimacodota: I think it would make more sense to add  sentry instead

dimacodota: but for now, write it to log, and add spere case for TimeoutException

amircodota: Hmmm.... What do we do in case of timeout? Just return null? What will happen when we send the next request? I'm pretty sure stuff will break if we send a request before receiving a response.

Were you able to reproduce it? How?

boaz-codota:
Added tests & refactored & some minor bug fixes (I'll write in the changes) @dimacodota

src/main/java/com/tabnine/TabNineProcess.java
```diff
@@ -2,37 +2,62 @@
...
     }
 
-    void startTabNine() throws IOException {
-        if (this.proc != null) {
-            this.proc.destroy();
-            this.proc = null;
+    public <T> T request(Request<T> request) {
+        Gson gson = new GsonBuilder().create();
+        Map<String, Object> jsonObject = new HashMap<>();
+        jsonObject.put("version", BINARY_VERSION);
```
Eransho, line 28: I realize it is not new, but why are we passing "2.0.2" as version property?

boaz-codota: This is the binary's protocol version

boaz-codota: I'll rename it

Eransho:
@boaz-codota hard to track on all the refactors. Is there any logic change in the refacor code or did you just move the original block(s) of code as is?

boaz-codota:
I've done some intensive refactoring & wrote extensive integration tests for the areas I touched.
Some improvements:
    * Restarts are offloaded to thread pool
    * Handles timeout properly
    * Handles bad results properly
        * Includes inner binary timeouts
    * Exponential backoff restart mechanism
        * To allow us to implement fixes without ever stopping.

PR #44, opened 2020-09-23
fix popup behavior

avnerbarr:


avnerbarr:
@amircodota @boaz-codota I think that the default merge behaviour for this repo is merge-commit (or possibly locally my defaults are screwed up) not merge-rebase causing the previous PR to include Boaz commits from rebased master so I replaced the previous PR with this PR  (will look into a fix in the meantime)

replaces
https://github.com/codota/tabnine-intellij/pull/40

PR #48, opened 2020-09-24
have different Config implementations for different channels

amircodota:


boaz-codota:
@amircodota I have no idea what this suppose to do

amircodota:
@boaz-codota Ths idea is to have a Config class with constants for each of the supported channels - alpha, beta, production, and include the right Config.java based on the given channel.
This way, we can have a different configuration based on the channel we are compiling for, so for example, we can know which channel we are on, what url to download the binary from etc.

boaz-codota:
> @boaz-codota Ths idea is to have a Config class with constants for each of the supported channels - alpha, beta, production, and include the right Config.java based on the given channel.
> This way, we can have a different configuration based on the channel we are compiling for, so for example, we can know which channel we are on, what url to download the binary from etc.

Is it a prep for the future? (If so, please note that "premature optimisation is the root of all evil")

amircodota:
@boaz-codota hehe... no, I use it to pass the clientChannel metadata field to the binary

src/main/java/com/tabnine/binary/BinaryRun.java
```diff
@@ -28,7 +29,8 @@ public BinaryRun(BinaryVersionFetcher binaryFetcher) {
             List<String> metadata = new ArrayList<>(asList(
                     "--client-metadata",
                     "pluginVersion=" + Utils.getPluginVersion(),
-                    "clientIsUltimate=" + PlatformUtils.isIdeaUltimate()
+                    "clientIsUltimate=" + PlatformUtils.isIdeaUltimate(),
+                    "clientChannel=" + Config.CHANNEL
```
boaz-codota, line 33: This is high risk change. The binary won't run if he does not recognise that parameter

boaz-codota (approved):

PR #53, opened 2020-10-06
Implement SetState

boaz-codota:


src/main/java/com/tabnine/contracts/AutocompleteRequest.java
```diff
@@ -1,28 +1,36 @@
...
+
+public class AutocompleteRequest implements BinaryRequest<Map<String, Object>, AutocompleteResponse> {
     public String before;
     public String after;
     public String filename;
-    public boolean region_includes_beginning;
-    public boolean region_includes_end;
-    public int max_num_results;
-    public Integer correlation_id;
+    @SerializedName(value = "region_includes_beginning")
```
dimacodota, line 16: nice 👍

src/main/java/com/tabnine/DependencyContainer.java
```diff
@@ -2,10 +2,33 @@
...
+
+            GATEWAY_INSTANCE.init();
+        }
+
+        return GATEWAY_INSTANCE;
+    }
+
+    public static synchronized TabNineLookupListener singletonOfTabNineLookupListener() {
+        if (LOOKUP_LISTENER_INSTANCE == null) {
+            LOOKUP_LISTENER_INSTANCE = new TabNineLookupListener();
```
dimacodota, line 26: why is it singleton? what state does it manages?

dimacodota (approved):
LGTM, had minor questin

src/main/java/com/tabnine/DependencyContainer.java
boaz-codota: Well they way IntelliJ does selections is weird, as it requires me to register the selection listener upon a completion suggestion (we did the same in Codota & I searched the web for how to do it), so to make sure that I won't register/initiate `TabNineLookupListener` multiple times I make it a singleton and removes it from the selection listener list before adding it again. At:
https://github.com/codota/tabnine-intellij/pull/53/files#diff-04d46f64b334bd2109e21f7780c6176fR133

dimacodota: 👍

PR #57, opened 2020-10-08
Use binary api for disable & uninstall

boaz-codota:


src/main/java/com/tabnine/selections/BinaryRequest.java
```diff
@@ -1,9 +1,14 @@
 package com.tabnine.selections;
 
+import com.tabnine.binary.exceptions.TabNineInvalidResponseException;
 import org.jetbrains.annotations.NotNull;
 
-public interface BinaryRequest<T, R extends BinaryResponse> {
+public interface BinaryRequest<R extends BinaryResponse> {
     Class<R> response();
-    T serialize(int correlationId);
+    Object serialize(int correlationId);
```
dimacodota, line 8: why?

dimacodota (commented):
see my question

src/main/java/com/tabnine/selections/BinaryRequest.java
boaz-codota: Because it is being serialized with Gson anyhow, so no need for a generic here. The generic enforces absolutely nothing....

dimacodota (approved):
LGTM

PR #60, opened 2020-10-11
Fix possible synchronized bug

boaz-codota:


boaz-codota:
@dimacodota I saw I had some restarts because of correlation id sync issues. I think this was the problem.

amircodota (approved):
LGTM

PR #62, opened 2020-10-21
Better concurrency

boaz-codota:


src/main/java/com/tabnine/TabNineCompletionContributor.java
```diff
@@ -32,6 +35,10 @@
 public class TabNineCompletionContributor extends CompletionContributor {
     private final TabNineGateway process = DependencyContainer.singletonOfTabNineGateway();
 
+    private int consecutiveRestarts = 0;
```
dimacodota, line 38: why not AtomicInteger ?

dimacodota (commented):
see my questions

src/main/java/com/tabnine/TabNineCompletionContributor.java
```diff
@@ -110,14 +123,28 @@ private AutocompleteResponse retrieveCompletions(CompletionParameters parameters
                 req.regionIncludesEnd = (end == doc.getTextLength());
 
                 try {
-                    return AppExecutorUtil.getAppExecutorService().submit(() -> this.process.request(req))
-                            .get(COMPLETION_TIME_THRESHOLD, TimeUnit.MILLISECONDS);
+                    Future<AutocompleteResponse> request = AppExecutorUtil.getAppExecutorService().submit(() -> this.process.request(req));
+
+                    activeRequests.add(request);
```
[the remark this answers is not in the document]
boaz-codota: hmm could be cleaned after the request was completed. How would you do that? run a background thread that cleans finished tasks?

src/main/java/com/tabnine/TabNineCompletionContributor.java
boaz-codota: No need, it is not multithreaded

boaz-codota: If it was multithreaded, IntelliJ wasn't hang when the response didn't arrive (that is why we needed to implement timeouts...)

src/main/java/com/tabnine/TabNineCompletionContributor.java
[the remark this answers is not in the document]
dimacodota: why not remove it after the ` this.process.request(req)` in the submit?

src/main/java/com/tabnine/TabNineCompletionContributor.java
dimacodota: oh, I thought you're modifying it in the submit... nevermind

boaz-codota:
Looks like this:
image

dimacodota (approved):
LGTM

PR #68, opened 2020-11-03
support bundles

avnerbarr:


avnerbarr:
we support bootstrapper bundles by adding a small bundle support layer in the plugin.
the support layer will download a version containing the bootstrapper and always start versions >= to that version (we have an API for bundle version containing the bootstrapper) and we assume that going forward that is correct

src/main/java/com/tabnine/binary/fetch/BinaryVersionFetcher.java
```diff
@@ -27,6 +28,12 @@ public BinaryVersionFetcher(LocalBinaryVersions localBinaryVersions, BinaryRemot
      * @throws SecurityException, NoExistingBinaryException if something went wrong
      */
     public String fetchBinary() throws NoValidBinaryToRunException {
+        Optional<BinaryVersion> bootstrappedVersion = BootstrapperSupport.bootstrap_version(localBinaryVersions, binaryRemoteSource, binaryDownloader);
+        if (bootstrappedVersion.isPresent()) {
+            Logger.getInstance(getClass()).info(format("found local bootstrapped version %s", bootstrappedVersion.get()));
+            return bootstrappedVersion.get().getVersionFullPath();
+        }
+        Logger.getInstance(getClass()).warn("couldn't get bootstrapped version. fallback to legacy code!");
```
boaz-codota, line 36: extract to a method in a different class

boaz-codota, line 36: Warning is a bit harsh here don't you think?W

src/main/java/com/tabnine/binary/fetch/BinaryVersionFetcher.java
```diff
@@ -27,6 +28,12 @@ public BinaryVersionFetcher(LocalBinaryVersions localBinaryVersions, BinaryRemot
      * @throws SecurityException, NoExistingBinaryException if something went wrong
      */
     public String fetchBinary() throws NoValidBinaryToRunException {
+        Optional<BinaryVersion> bootstrappedVersion = BootstrapperSupport.bootstrap_version(localBinaryVersions, binaryRemoteSource, binaryDownloader);
+        if (bootstrappedVersion.isPresent()) {
+            Logger.getInstance(getClass()).info(format("found local bootstrapped version %s", bootstrappedVersion.get()));
```
boaz-codota, line 33: debug level instead of info level

src/main/java/com/tabnine/binary/fetch/BinaryVersionFetcher.java
```diff
@@ -27,6 +28,12 @@ public BinaryVersionFetcher(LocalBinaryVersions localBinaryVersions, BinaryRemot
      * @throws SecurityException, NoExistingBinaryException if something went wrong
      */
     public String fetchBinary() throws NoValidBinaryToRunException {
+        Optional<BinaryVersion> bootstrappedVersion = BootstrapperSupport.bootstrap_version(localBinaryVersions, binaryRemoteSource, binaryDownloader);
```
boaz-codota, line 31: I think we can leverage `orElseGet` here after the refactor

boaz-codota, line 31: This being static is interfering with [Dependency Inversion Principle](https://en.wikipedia.org/wiki/Dependency_inversion_principle). So make it a class and consume an instance through the constructor. Exceptions are everywhere here.

src/main/java/com/tabnine/binary/fetch/BootstrapperSupport.java
```diff
@@ -0,0 +1,73 @@
...
+import com.intellij.util.text.SemVer;
+import com.tabnine.general.StaticConfig;
+
+import java.util.List;
+import java.util.Optional;
+import java.util.prefs.Preferences;
+
+import static java.lang.String.format;
+
+public class BootstrapperSupport {
```
boaz-codota, line 13: Nothing here needs to be static

src/main/java/com/tabnine/binary/fetch/BootstrapperSupport.java
```diff
@@ -0,0 +1,73 @@
...
+    static Optional<BinaryVersion> bootstrap_version(LocalBinaryVersions localBinaryVersions, BinaryRemoteSource binaryRemoteSource, BinaryDownloader binaryDownloader) {
```
boaz-codota, line 14: Receive dependencies through the constructor.

boaz-codota, line 14: Java naming convention for method names is without underscores, like so: `bootstrapVersion`. But I would like to suggest you use a better name for it that include some verb like `findVersion`

src/main/java/com/tabnine/binary/fetch/BootstrapperSupport.java
boaz-codota, line 13: This name is too generic and encourages others to add more methods to this class. How about `BootstraperVersions`?

src/main/java/com/tabnine/binary/fetch/BootstrapperSupport.java
```diff
@@ -0,0 +1,73 @@
...
+
+    }
+
+    static String BOOTSTRAPPED_VERSION_KEY = "bootstrapped version";
+
+    static Preferences get_prefs() {
+        return Preferences.userNodeForPackage(BootstrapperSupport.class);
+    }
+
+    /// returns the minimal known version supporting the bootstrapper
```
boaz-codota, line 62: no need for so much comments. Please remove them all.

Only acceptable comments are when:
* Some optimisation took place, sacrificing readability.
* We took a shortcut and we would like encourage people to improve in the future.

src/main/java/com/tabnine/general/StaticConfig.java
```diff
@@ -42,15 +42,26 @@
 
     public static final Icon ICON = IconLoader.findIcon("/icons/tabnine-icon-13px.png");
 
+
     public static String getServerUrl() {
         return Optional.ofNullable(System.getProperty(REMOTE_BASE_URL_PROPERTY)).orElse("https://update.tabnine.com");
     }
 
+    // https://update.tabnine.com/bundles/version
```
boaz-codota, line 50: remove

src/main/java/com/tabnine/general/StaticConfig.java
```diff
@@ -42,15 +42,26 @@
...
+    }
+
     @NotNull
     public static String getTabNineVersionUrl() {
         return Optional.ofNullable(System.getProperty(REMOTE_VERSION_URL_PROPERTY)).orElse(getServerUrl() + "/version");
     }
 
+    @NotNull
+    public static String getTabNineBundleVersionUrl() {
+        return Optional.ofNullable(System.getProperty(REMOTE_VERSION_URL_PROPERTY)).orElse(getBundleServerUrl() + "/version");
```
boaz-codota, line 62: This will resolve `https://update.tabnine.com/bundles/version`, is that the intended url?

src/main/java/com/tabnine/binary/fetch/BootstrapperSupport.java
```diff
@@ -0,0 +1,73 @@
...
+        }
+
+        return binaryVersion;
+
+    }
+
+    static String BOOTSTRAPPED_VERSION_KEY = "bootstrapped version";
+
+    static Preferences get_prefs() {
+        return Preferences.userNodeForPackage(BootstrapperSupport.class);
```
boaz-codota, line 59: I have no idea what this does, care to explain to me? 🙏🏻

src/main/java/com/tabnine/binary/fetch/BootstrapperSupport.java
```diff
@@ -0,0 +1,73 @@
...
+        } else {
+            // will fallback to the legacy code in the BinaryVersionFetcher
+            Logger.getInstance(BootstrapperSupport.class).warn("couldn't download bootstrapper bundle!");
+        }
+
+        return binaryVersion;
+
+    }
+
+    static String BOOTSTRAPPED_VERSION_KEY = "bootstrapped version";
```
boaz-codota, line 56: Consts in Java should be of form:
```
    private static final String BOOTSTRAPPED_VERSION_KEY = "bootstrapped version";
```

src/test/java/com/tabnine/binary/BinaryVersionFetcherTests.java
```diff
@@ -30,26 +34,36 @@
...
+        } catch (BackingStoreException e) {
+            e.printStackTrace();
+        }
+    }
     @Test
     public void givenBetaVersionThatAvailableLocallyWhenFetchBinaryThenBetaVersionReturned() throws Exception {
         when(localBinaryVersions.listExisting()).thenReturn(aVersions());
         when(binaryRemoteSource.existingLocalBetaVersion(aVersions())).thenReturn(Optional.of(new BinaryVersion(BETA_VERSION)));
-
+        when(binaryRemoteSource.fetchPreferredVersion(StaticConfig.getTabNineBundleVersionUrl())).thenReturn(Optional.empty());
```
boaz-codota, line 50: this line repeats on all tests so please extract it to the `@Before`

src/test/java/com/tabnine/binary/BinaryVersionFetcherTests.java
```diff
@@ -30,26 +34,36 @@
     @InjectMocks
     private BinaryVersionFetcher binaryVersionFetcher;
 
+    @Before
+    public void clearBootstrapperPreferences() {
+        Preferences preferences = Preferences.userNodeForPackage(BootstrapperSupport.class);
+        try {
+            preferences.clear();
+        } catch (BackingStoreException e) {
+            e.printStackTrace();
```
boaz-codota, line 43: better throw that exception so the tests will fail.

src/test/java/com/tabnine/binary/BinaryVersionFetcherTests.java
```diff
@@ -2,13 +2,17 @@
```
boaz-codota, line 2: You did not add tests for the code you added.

src/main/java/com/tabnine/binary/fetch/BootstrapperSupport.java
```diff
@@ -0,0 +1,73 @@
...
+    static Optional<SemVer> minimalBootstrappedVersion() {
+        String nonExistantSentinal = "non existent sentinel";
+        // annoyingly requires a default value parameter
+        String minimalBootstrappedVersion = get_prefs().get(BOOTSTRAPPED_VERSION_KEY, nonExistantSentinal);
+        if (!minimalBootstrappedVersion.equals(nonExistantSentinal)) {
+            SemVer thisVersion = SemVer.parseFromText(minimalBootstrappedVersion);
+            return Optional.ofNullable(thisVersion);
+        }
+        return Optional.empty();
+    }
```
boaz-codota, line 72: Could be written as:
```
return Optional.ofNullable(get_prefs().get(BOOTSTRAPPED_VERSION_KEY)).map(SemVer::parseFromText).filter(v => v != null);
```

src/main/java/com/tabnine/binary/fetch/BootstrapperSupport.java
```diff
@@ -0,0 +1,73 @@
...
+
+        if (thisVersion.isPresent()) {
+            // versions list is returned reverse ordered by semver
+            // take first which has a version >= thisVersion
+            // (since bootstrapper can download newer versions or downgrade but we let the bootstrapper handle that)
+            // "for" since parseFromText returns nullable and don't want to deal with ugliness in map/filter
+            List<BinaryVersion> versions = localBinaryVersions.listExisting();
+            for (BinaryVersion v : versions) {
+                SemVer ver = SemVer.parseFromText(v.getVersion());
+                if (ver.compareTo(thisVersion.get()) >= 0) {
```
boaz-codota, line 29: give it a name

src/main/java/com/tabnine/binary/fetch/BootstrapperSupport.java
```diff
@@ -0,0 +1,73 @@
...
+
+        Optional<SemVer> thisVersion = minimalBootstrappedVersion();
+
+        if (thisVersion.isPresent()) {
+            // versions list is returned reverse ordered by semver
+            // take first which has a version >= thisVersion
+            // (since bootstrapper can download newer versions or downgrade but we let the bootstrapper handle that)
+            // "for" since parseFromText returns nullable and don't want to deal with ugliness in map/filter
+            List<BinaryVersion> versions = localBinaryVersions.listExisting();
+            for (BinaryVersion v : versions) {
```
boaz-codota, line 27: use `stream`

src/main/java/com/tabnine/binary/fetch/BootstrapperSupport.java
```diff
@@ -0,0 +1,73 @@
...
+
+public class BootstrapperSupport {
+    static Optional<BinaryVersion> bootstrap_version(LocalBinaryVersions localBinaryVersions, BinaryRemoteSource binaryRemoteSource, BinaryDownloader binaryDownloader) {
+        // bootstrapper can change versions under the hood
+        // plugin must ensure it is running a version containing the bootstrapper but not override any versions that the bootstrapper may have downloaded on its own
+        // we persist the first known bootstrapper version and choose accordingly
+
+        Optional<SemVer> thisVersion = minimalBootstrappedVersion();
+
+        if (thisVersion.isPresent()) {
```
boaz-codota, line 21: prefer map & orElseGet, this is a general comment for when using optional

src/main/java/com/tabnine/binary/fetch/BootstrapperSupport.java
```diff
@@ -0,0 +1,73 @@
+package com.tabnine.binary.fetch;
```
boaz-codota, line 1: Add private/public access levels to all methods and members. No need to make anything package-protected.

src/main/java/com/tabnine/binary/fetch/BootstrapperSupport.java
```diff
@@ -0,0 +1,73 @@
...
+            String version = binaryVersion.get().getVersion();
+            // we save this as the "minimal bootstrap supported version"
+            get_prefs().put(BOOTSTRAPPED_VERSION_KEY, version);
+        } else {
+            // will fallback to the legacy code in the BinaryVersionFetcher
+            Logger.getInstance(BootstrapperSupport.class).warn("couldn't download bootstrapper bundle!");
+        }
+
+        return binaryVersion;
+
```
boaz-codota, line 53: Extract method

src/main/java/com/tabnine/binary/fetch/BootstrapperSupport.java
```diff
@@ -0,0 +1,73 @@
...
+    static String BOOTSTRAPPED_VERSION_KEY = "bootstrapped version";
+
+    static Preferences get_prefs() {
+        return Preferences.userNodeForPackage(BootstrapperSupport.class);
+    }
+
+    /// returns the minimal known version supporting the bootstrapper
+    static Optional<SemVer> minimalBootstrappedVersion() {
+        String nonExistantSentinal = "non existent sentinel";
```
boaz-codota, line 64: No need to match a random string when you can match null

src/main/java/com/tabnine/binary/fetch/BootstrapperSupport.java
```diff
@@ -0,0 +1,73 @@
...
+
+        Optional<BinaryVersion> binaryVersion = binaryRemoteSource
+                                                    .fetchPreferredVersion(StaticConfig.getTabNineBundleVersionUrl())
+                                                    .flatMap(binaryDownloader::downloadBinary);
+
+        if (binaryVersion.isPresent()) {
+            Logger.getInstance(BootstrapperSupport.class).info(format("downloaded bootstrap bundle %s", binaryVersion.get()));
+            String version = binaryVersion.get().getVersion();
+            // we save this as the "minimal bootstrap supported version"
+            get_prefs().put(BOOTSTRAPPED_VERSION_KEY, version);
```
boaz-codota, line 46: Is this some sort of a caching mechanism? It persists on disk or only in memory? Why do we need to cache this?

boaz-codota (changes requested):

boaz-codota:
> we support bootstrapper bundles by adding a small bundle support layer in the plugin.
> the support layer will download a version containing the bootstrapper and always start versions >= to that version (we have an API for bundle version containing the bootstrapper) and we assume that going forward that is correct

Does need to be done on any IDE or only for IntelliJ?

src/main/java/com/tabnine/binary/fetch/BinaryVersionFetcher.java
avnerbarr: it is in a different class

avnerbarr: the server should always return a version unless problems

src/main/java/com/tabnine/binary/fetch/BinaryVersionFetcher.java
avnerbarr: the other log in the class is info for the same purpose

src/main/java/com/tabnine/binary/fetch/BinaryVersionFetcher.java
avnerbarr: the orelse part is all the existing code

avnerbarr: this is all temporary until we are going to remove the other code. . I don't think it is worth the time effort as the bootstrapper support code will be inlined to the existing function

src/main/java/com/tabnine/binary/fetch/BootstrapperSupport.java
avnerbarr: obviously. there is additional logic here and I don't like the functional style that would be nested

src/main/java/com/tabnine/binary/fetch/BootstrapperSupport.java
avnerbarr: why?

src/main/java/com/tabnine/binary/fetch/BootstrapperSupport.java
avnerbarr: I'll try.

src/main/java/com/tabnine/binary/fetch/BootstrapperSupport.java
avnerbarr: disk . we want to download only once and then delegate the version management to the bootstrapper

src/main/java/com/tabnine/binary/fetch/BootstrapperSupport.java
avnerbarr: its a piure function. I don't want to create an instance for this

src/main/java/com/tabnine/binary/fetch/BootstrapperSupport.java
avnerbarr: why? what is the advantage? this just a once off funciton and no need to construct it

src/main/java/com/tabnine/general/StaticConfig.java
avnerbarr: yep

src/main/java/com/tabnine/binary/fetch/BinaryVersionFetcher.java
boaz-codota: There is no such thing as a temporary code. Everything in master should be addressed as production code.

src/main/java/com/tabnine/binary/fetch/BootstrapperSupport.java
boaz-codota: because this method is more than ~10 lines long, making it a story instead of a poem.

src/main/java/com/tabnine/binary/fetch/BootstrapperSupport.java
boaz-codota: Where on disk is this cached?

boaz-codota: Please add a test that verifies it persists on disk

src/main/java/com/tabnine/binary/fetch/BootstrapperSupport.java
boaz-codota: Yep, I've found those docs as well, I truly asking for help as I'm not familiar with it. Some explanation on how it works (or some examples) would make me grateful.

src/main/java/com/tabnine/binary/fetch/BootstrapperSupport.java
```diff
@@ -0,0 +1,73 @@
...
+
+    static String BOOTSTRAPPED_VERSION_KEY = "bootstrapped version";
+
+    static Preferences get_prefs() {
+        return Preferences.userNodeForPackage(BootstrapperSupport.class);
+    }
+
+    /// returns the minimal known version supporting the bootstrapper
+    static Optional<SemVer> minimalBootstrappedVersion() {
+        String minimalBootstrappedVersion = get_prefs().get(BOOTSTRAPPED_VERSION_KEY, null);
```
boaz-codota, line 66: use Optional.ofNullable and then map

boaz-codota:
```java
return Optional.ofNullable(get_prefs().get(BOOTSTRAPPED_VERSION_KEY)).map(SemVer::parseFromText)
```

boaz-codota (changes requested):

src/main/java/com/tabnine/binary/fetch/BootstrapperSupport.java
avnerbarr: rust habit :)

src/main/java/com/tabnine/binary/fetch/BootstrapperSupport.java
avnerbarr: there is one version

src/main/java/com/tabnine/binary/fetch/BootstrapperSupport.java
avnerbarr:
```A node in a hierarchical collection of preference data.  This class
 * allows applications to store and retrieve user and system
 * preference and configuration data.  This data is stored
 * persistently in an implementation-dependent backing store.  Typical
 * implementations include flat files, OS-specific registries,
 * directory servers and SQL databases.  The user of this class needn't
 * be concerned with details of the backing store.
```

src/main/java/com/tabnine/binary/fetch/BootstrapperSupport.java
avnerbarr: simple key value

src/main/java/com/tabnine/binary/fetch/BootstrapperSupport.java
```diff
@@ -0,0 +1,47 @@
...
+// once out of beta/alpha all of this code can replace the existing fetchBinary()
+public class BootstrapperSupport {
+    static Optional<BinaryVersion> bootstrapVersion(LocalBinaryVersions localBinaryVersions, BinaryRemoteSource binaryRemoteSource, BundleDownloader bundleDownloader) {
+        Optional<BinaryVersion> localBootstrapVersion = locateLocalBootstrapSupportedVersion(localBinaryVersions);
+        return localBootstrapVersion.isPresent() ? localBootstrapVersion : downloadRemoteVersion(binaryRemoteSource, bundleDownloader);
+    }
+
+    private static final String BOOTSTRAPPED_VERSION_KEY = "bootstrapped version";
+
+    private static Preferences get_prefs() {
```
dimacodota, line 19: rust? remove the underscore

src/test/java/com/tabnine/binary/BootstrapperSupportTests.java
```diff
@@ -0,0 +1,80 @@
...
+    private BinaryRemoteSource binaryRemoteSource;
+    @Mock
+    private BundleDownloader bundleDownloader;
+
+    @InjectMocks
+    private BinaryVersionFetcher binaryVersionFetcher;
+
+    @BeforeEach
+    public void setUp() throws BackingStoreException {
+        Preferences preferences = Preferences.userNodeForPackage(BootstrapperSupport.class);
```
dimacodota, line 35: see my comment  regarding the Preferences.userNodeForPackage impl coupling

dimacodota (approved):
Had some minor comment, they are not show stopper, but a general paradigm choice alignment. we have chosen to follow the OOP principles here (DI, IOC, ctor injection, no new, no statics, etc.), as an enabler for better testing robustness infra.

src/main/java/com/tabnine/binary/fetch/BundleDownloader.java
```diff
@@ -0,0 +1,69 @@
...
+            }
+            fos.close();
+            zipEntry = zis.getNextEntry();
+            newFile.setExecutable(true);
+        }
+        zis.closeEntry();
+        zis.close();
+    }
+
+    private static File newFile(File destinationDir, ZipEntry zipEntry) throws IOException {
```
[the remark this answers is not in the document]
avnerbarr: From my experience, functions not acting on the "self" instance should be class functions (or static) depends on the language syntax.

src/main/java/com/tabnine/binary/fetch/GeneralDownloader.java
```diff
@@ -0,0 +1,48 @@
...
+import java.nio.file.Paths;
+import java.nio.file.StandardCopyOption;
+import java.util.UUID;
+
+import static com.tabnine.general.StaticConfig.BINARY_READ_TIMEOUT;
+import static com.tabnine.general.StaticConfig.REMOTE_CONNECTION_TIMEOUT;
+import static java.lang.String.format;
+
+public class GeneralDownloader {
+    static boolean download(String urlString, String destination, DownloadValidator validator) {
```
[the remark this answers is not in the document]
avnerbarr: there is no need for an instance

src/main/java/com/tabnine/binary/fetch/BootstrapperSupport.java
avnerbarr: :)

src/test/java/com/tabnine/binary/BootstrapperSupportTests.java
avnerbarr: where are the comments?

src/main/java/com/tabnine/binary/fetch/GeneralDownloader.java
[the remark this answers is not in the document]
avnerbarr: I can inject it

src/test/java/com/tabnine/binary/BootstrapperSupportTests.java
dimacodota: https://github.com/codota/tabnine-intellij/pull/68#discussion_r517401362

src/main/java/com/tabnine/binary/fetch/BootstrapperSupport.java
```diff
@@ -0,0 +1,47 @@
...
+        getPrefs().put(BOOTSTRAPPED_VERSION_KEY, version.getVersion());
+        return version;
+    }
+
+    private static Optional<SemVer> minimalBootstrappedVersion() {
+        return Optional.ofNullable(getPrefs().get(BOOTSTRAPPED_VERSION_KEY, null)).map(SemVer::parseFromText);
+    }
+
+    private static Optional<BinaryVersion> locateLocalBootstrapSupportedVersion(LocalBinaryVersions localBinaryVersions) {
+        return minimalBootstrappedVersion().flatMap(
```
amircodota, line 33: shouldn't we always take the most recent one? Maybe I'm mssing something, but `findFirst` indicates that we can select pretty much any of them, no?

src/test/java/com/tabnine/binary/BootstrapperSupportTests.java
```diff
@@ -0,0 +1,80 @@
...
+        assertThat(binaryVersionFetcher.fetchBinary(), equalTo(versionFullPath("66.66.66")));
+    }
+    @Test
+    public void testWhenBootstrappedVersionFailedToDownloadWillFallbackToLocalVersion() throws Exception {
+        when(localBinaryVersions.listExisting()).thenReturn(aVersions());
+        when(binaryRemoteSource.fetchPreferredVersion(StaticConfig.getTabNineBundleVersionUrl())).thenReturn(Optional.of("7.7.7"));
+        when(binaryRemoteSource.fetchPreferredVersion()).thenReturn(Optional.of(PREFERRED_VERSION));
+        when(bundleDownloader.downloadAndExtractBundle("7.7.7")).thenReturn(Optional.empty());
+        assertThat(binaryVersionFetcher.fetchBinary(), equalTo(versionFullPath(PREFERRED_VERSION)));
+    }
```
amircodota, line 79: I would add a test here to make sure that when multiple versions exist, it will pick the latest

amircodota (commented):
see my comments

src/main/java/com/tabnine/binary/fetch/BootstrapperSupport.java
avnerbarr: it will

src/test/java/com/tabnine/binary/BootstrapperSupportTests.java
avnerbarr: localBinaryVersions.listExisting() returns them ordered by latest

src/main/java/com/tabnine/binary/fetch/BootstrapperSupport.java
avnerbarr: `.filter(v -> SemVer.parseFromText(v.getVersion()).compareTo(version) >= 0)` and then taking the first (which is >= version)

amircodota: :+1:

amircodota: Consider renaming the method or something, because at least for me, it was not obvious form the code

amircodota (approved):
LGTM

PR #70, opened 2020-11-08
log file path support for IJ

amircodota:


build.gradle
```diff
@@ -60,3 +60,7 @@ tasks.create("currentVersion") {
         println version
     }
 }
+
+/*tasks.getByName("runIde").configure {
+    systemProperty("TABNINE_LOG_FILE_PATH", "/home/amir/t9-ij.log")
+}*/
```
boaz-codota, line 66: Dont leave commented code

src/main/java/com/tabnine/binary/BinaryRun.java
```diff
@@ -69,6 +71,12 @@ public Process reportUninstall(String... additionalMetadata) throws NoValidBinar
                 metadata.add("clientVersion=" + cmdSanitize(applicationInfo.getFullVersion()));
                 metadata.add("clientApiVersion=" + cmdSanitize(applicationInfo.getApiVersion()));
             }
+
+            StaticConfig.getLogFilePath().ifPresent(v -> {
```
boaz-codota, line 75: Add static import

boaz-codota (approved):
Awesome & simple implementation. Some minor cosmetics I'd like you to change

build.gradle
amircodota: This is the example of how to use this parameter in debug. 
It is not immediately obvious.
I'll add a comment explaining

amircodota: Actually, I did something different.

I added a `logFilePath` gradle property that sets this value, so that you can add `-PlogFilePath={path)` when running form gradle to set it

boaz-codota: @amircodota maybe worth documenting this somewhere?

amircodota: where?

PR #71, opened 2020-11-08
Fix concurrency issues

boaz-codota:


boaz-codota:
Can't think on how to test this... The `volatile` is the key to make this work.

dimacodota (approved):
LGTM

PR #75, opened 2020-11-17
remove annoying message in log

amircodota:


boazberman:
@amircodota please add test

amircodota:
I did

boaz-codota:
> I did

I think that the test you added would have pass without the code changes... Hence it is a bad test (if I'm wrong I'd love to know)

amircodota:
You are not wrong.

To do what you are asking, I'd either have to abstract away the logger so I can mock it in a test and make sure nothing is written to it, or abstract the ProcessBuilder, and make sure nothing is being run.

These abstractions do not exist, and adding this complexity for the sake of testing this small fix is just not worth it, IMO, both in terms of the effort required and in terms of the added complexity.

Wouldn't you agree?

boaz-codota:
> You are not wrong.
> 
> To do what you are asking, I'd either have to abstract away the logger so I can mock it in a test and make sure nothing is written to it, or abstract the ProcessBuilder, and make sure nothing is being run.
> 
> These abstractions do not exist, and adding this complexity for the sake of testing this small fix is just not worth it, IMO, both in terms of the effort required and in terms of the added complexity.
> 
> Wouldn't you agree?

I disagree... You can write test that includes temp folder (there is a rule for that in junit) and test using that temp folder and mock data...

amircodota:
And that test would have failed before and pass now? How?

boaz-codota:
> And that test would have failed before and pass now? How?

Because of an unexisting file? Listen if we can't test this change, it does nothing.

amircodota:
The things is it behaved correctly before, except that it wrote a nasty error to the log, which I am trying to avoid.

boaz-codota:
> The things is it behaved correctly before, except that it wrote a nasty error to the log, which I am trying to avoid.

Can't we test the log?

amircodota:
Ofcourse we can. It requires adding the log as a dependency, and then mock it.
We never do this for any of our tests, it will add an unnecessary complexity and is not worth the hassle, in my opinion

PR #97, opened 2021-01-05
chnaged Manual Installation inside VSCode section of readme file. Now…

YotamFromCodota:
… includes specs for Windows also

### DIFF
--- a/src/main/resources/META-INF/plugin.xml
+++ b/src/main/resources/META-INF/plugin.xml
@@ -72,12 +72,12 @@
 
 <h2 id="-installation-instructions-"><strong>Installation Instructions</strong></h2>
 <ul>
-    <li>Press Cmd+,to open the Settings/Preferences dialog</li>
-    <li>Search for TabNine in the Marketplace tab and press Install</li>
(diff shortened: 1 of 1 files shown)


src/main/resources/META-INF/plugin.xml
```diff
@@ -102,7 +102,7 @@
 
 <h3 id="-usage-"><strong>Usage</strong></h3>
 <p>
-    After Tabnine’s installation, navigate to the Tabnine Settings page (‘Tabnine: Open Settings’ from the <a href="https://code.visualstudio.com/docs/getstarted/userinterface#_command-palette">Command Palette</a>) and verify that Tabnine local model is successfully loaded (as shown in the following screenshots)
+    After Tabnine’s installation, navigate to the Tabnine's plugin section (go to -> file -> settings -> plugins) and verify that Tabnine local model is successfully loaded.
```
boaz-codota, line 105: Replace the `'` with HTML compatible symobl, such as `&apos;`. See: https://www.w3.org/wiki/Common_HTML_entities_used_for_typography

PR #101, opened 2021-01-14
This was reported by a user in the below ticket:

shaiahr:
https://codota.zendesk.com/agent/tickets/2410
I couldn't reproduce so I'm not sure what's causing this and how to fix it. Meanwhile we'll just catch runtime exceptions when inserting a completion (so it won't jump to the user) and write more data the the log.

### DIFF
--- a/src/main/java/com/tabnine/TabNineCompletionContributor.java
+++ b/src/main/java/com/tabnine/TabNineCompletionContributor.java
@@ -2,6 +2,7 @@ package com.tabnine;
 
 import com.intellij.codeInsight.completion.*;
 import com.intellij.codeInsight.lookup.*;
+import com.intellij.openapi.diagnostic.Logger;
 import com.intellij.openapi.editor.Document;
 import com.intellij.openapi.util.TextRange;
 import com.tabnine.binary.requests.autocomplete.AutocompleteResponse;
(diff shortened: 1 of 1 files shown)


src/main/java/com/tabnine/TabNineCompletionContributor.java
```diff
@@ -97,8 +98,15 @@ private LookupElementBuilder createCompletion(CompletionParameters parameters, C
                 .withInsertHandler((context, item) -> {
                     int end = context.getTailOffset();
                     TabNineCompletion lookupElement = (TabNineCompletion) item.getObject();
-                    context.getDocument().insertString(end + lookupElement.oldSuffix.length(), lookupElement.newSuffix);
-                    context.getDocument().deleteString(end, end + lookupElement.oldSuffix.length());
+                    try {
+                        if (true) throw new IndexOutOfBoundsException("shai: some error");
```
boaz-codota, line 102: ??

boaz-codota (commented):
I'm not sure what benefits we will get from this. Maybe the user enabled multiline completions and we have issues with those in IntelliJ?

boaz-codota (changes requested):
`if(true)` that throws an exception is not something we would like in production I believe

src/main/java/com/tabnine/TabNineCompletionContributor.java
boaz-codota: Also, please provide better message to the exception

shaiahr: Pushed this line by mistake.

PR #113, opened 2021-02-16
Init users of BinaryRun on appStarting to prevent error: "Should be c…

shaiahr:
…alled at least in the state CONFIGURATION_STORE_INITIALIZED, the current state is: COMPONENTS_REGISTERED" since BinaryRun calls ApplicationInfo.getInstance().


boaz-codota (approved):
🚀

PR #114, opened 2021-02-16
Fix no destroy on restart & ktlint & Kotlin migration

boaz-codota:
Ref https://codota.atlassian.net/browse/DEV-2039


src/main/java/com/tabnine/binary/BinaryProcessGateway.java
```diff
@@ -36,4 +36,8 @@ public void writeRequest(String request) throws IOException {
     public boolean isDead() {
         return process == null || !process.isAlive();
     }
+
+    public void destroy() {
+        process.destroy();
```
shaiahr, line 41: process can't be null?
What happens if destroy throws an exception (don't know if that can happen)?

src/main/java/com/tabnine/lifecycle/BinaryNotificationsLifecycle.kt
```diff
@@ -11,52 +11,56 @@ import com.tabnine.binary.requests.notifications.NotificationOptions
...
-        }, BINARY_NOTIFICATION_POLLING_INTERVAL, BINARY_NOTIFICATION_POLLING_INTERVAL)
+            },
+            BINARY_NOTIFICATION_POLLING_INTERVAL, BINARY_NOTIFICATION_POLLING_INTERVAL
+        )
     }
 
-    private fun storageKey(id: String?) = "tabnine-notifications-shown-${id}"
-}
\ No newline at end of file
+    private fun storageKey(id: String?) = "tabnine-notifications-shown-$id"
```
shaiahr, line 65: Hard to tell if this is the only thing that changed here. I think it's better to separate format changes into a dedicated PR.
Also, this looks like a bug for something that was never tested? Did you test it now?

src/test/java/com/tabnine/binary/fetch/LocalBinaryVersionsTests.kt
```diff
@@ -0,0 +1,126 @@
...
+    private val originalHome = System.getProperty(USER_HOME_PATH_PROPERTY)
+
+    @BeforeEach
+    fun setup() {
+        System.setProperty(USER_HOME_PATH_PROPERTY, temporaryFolder.toString())
+    }
+
+    @AfterEach
+    fun teardown() {
+        System.setProperty(USER_HOME_PATH_PROPERTY, originalHome)
```
shaiahr, line 46: Shouldn't we also delete the temp folder here?

shaiahr (approved):
LGTM
See my questions about null and exception handling.

src/main/java/com/tabnine/binary/BinaryProcessGateway.java
boaz-codota: good catch

boaz-codota: Nope it can't. `init` is called

src/main/java/com/tabnine/lifecycle/BinaryNotificationsLifecycle.kt
boaz-codota: Nothing changed in that file, I only ran `ktlint` to format the whole files

src/test/java/com/tabnine/binary/fetch/LocalBinaryVersionsTests.kt
boaz-codota: JUnit takes care of that https://junit.org/junit5/docs/5.4.0/api/org/junit/jupiter/api/io/TempDir.html

src/main/java/com/tabnine/binary/BinaryProcessGateway.java
shaiahr: We check for null in isDead so is the check redundant there or missing here?

src/main/java/com/tabnine/lifecycle/BinaryNotificationsLifecycle.kt
shaiahr: Looks like the curly braces around the id were deleted.

src/test/java/com/tabnine/binary/fetch/LocalBinaryVersionsTests.kt
shaiahr: Cool.

PR #118, opened 2021-02-21
Use info messages background color for 'limit exceeded' inlay hover.

shaiahr:
Bug fix: Don't add the 'limit exceeded' inlay if one is already displayed at the current line.


dimacodota (approved):
lgtm

PR #120, opened 2021-02-23
Fix NullPointerException on engine process destroy

boaz-codota:
Ref: https://twitter.com/iyedoost/status/1364077269266948096?s=21
image


dimacodota (approved):
lgtm

src/main/java/com/tabnine/binary/BinaryProcessGateway.kt
```diff
@@ -0,0 +1,38 @@
+package com.tabnine.binary
+
+import com.tabnine.binary.exceptions.TabNineDeadException
+import java.io.BufferedReader
+import java.io.IOException
+import java.io.InputStreamReader
+import java.nio.charset.StandardCharsets
+
+open class BinaryProcessGateway {
+    private var process: Process? = null
```
shaiahr, line 10: Why should process be nullable? Won't it be better to create it in the constructor and throw some exception if it fails?

PR #121, opened 2021-02-24
Send the message id in 'Status Bar Promotion Shown' requests.

shaiahr:
Added missing fields to NotificationShownRequest, BinaryNotification, StatusBarPromotionBinaryResponse and StatusBarPromotionShownRequest.
Update status bar on selection (instead of just waiting for the poller).
Display status bar messages in all open project windows (instead of just the first opened one).

### DIFF
--- a/src/main/java/com/tabnine/lifecycle/BinaryPromotionStatusBarLifecycle.kt
+++ b/src/main/java/com/tabnine/lifecycle/BinaryPromotionStatusBarLifecycle.kt
@@ -1,61 +1,21 @@
 package com.tabnine.lifecycle
 
-import com.intellij.openapi.project.ProjectManager
-import com.intellij.openapi.wm.WindowManager
 import com.tabnine.binary.BinaryRequestFacade
-import com.tabnine.binary.requests.statusBar.StatusBarPromotionBinaryRequest
-import com.tabnine.binary.requests.statusBar.StatusBarPromotionShownRequest
 import com.tabnine.general.StaticConfig.BINARY_PROMOTION_POLLING_DELAY
 import com.tabnine.general.StaticConfig.BINARY_PROMOTION_POLLING_INTERVAL
-import com.tabnine.statusBar.StatusBarPromotionWidget
+import com.tabnine.statusBar.StatusBarUpdater
 import java.util.Timer
-import java.util.TimerTask
+import kotlin.concurrent.timerTask
 
 class BinaryPromotionStatusBarLifecycle(private val binaryRequestFacade: BinaryRequestFacade) {
+    private val statusBarUpdater = StatusBarUpdater(binaryRequestFacade)
+
     fun poll() {
         Timer().schedule(
-            object : TimerTask() {
-                override fun run() {
-                    val statusBarPromotionWidget = getPromotionWidget()
-
-                    val promotion = binaryRequestFacade.executeRequest(StatusBarPromotionBinaryRequest())
-
-                    if (promotion != null) {
-                        statusBarPromotionWidget?.isVisible = true
-                        statusBarPromotionWidget?.text = promotion.message
-                        statusBarPromotionWidget?.id = promotion.id
-                        statusBarPromotionWidget?.actions = promotion.actions
-                        statusBarPromotionWidget?.notificationType = promotion.notificationType
-
-                        binaryRequestFacade.executeRequest(StatusBarPromotionShownRequest(promotion.message ?: "undefined"))
-                    } else {
-                        clear(statusBarPromotionWidget)
-                    }
-                }
+            timerTask {
+                statusBarUpdater.requestStatusBarMessage()
             },
             BINARY_PROMOTION_POLLING_DELAY, BINARY_PROMOTION_POLLING_INTERVAL
         )
-    }
-
-    private fun getPromotionWidget(): StatusBarPromotionWidget.StatusBarPromotionComponent? {
-        val openProjects = ProjectManager.getInstance().openProjects
-
-        if (openProjects.isEmpty()) {
-            return null
-        }
-
-        val statusBar = WindowManager.getInstance()?.getStatusBar(openProjects[0])
-        val widget = statusBar?.getWidget(StatusBarPromotionWidget::class.java.name) ?: return null
-        val promotionWidget = widget as StatusBarPromotionWidget
-
-        return promotionWidget.component as StatusBarPromotionWidget.StatusBarPromotionComponent
-    }
-
-    private fun clear(statusBarPromotionWidget: StatusBarPromotionWidget.StatusBarPromotionComponent?) {
-        statusBarPromotionWidget?.text = null
-        statusBarPromotionWidget?.id = null
-        statusBarPromotionWidget?.isVisible = false
-        statusBarPromotionWidget?.actions = null
-        statusBarPromotionWidget?.notificationType = null
     }
 }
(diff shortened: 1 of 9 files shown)


src/main/java/com/tabnine/lifecycle/BinaryPromotionStatusBarLifecycle.kt
```diff
@@ -1,61 +1,21 @@
...
-                        statusBarPromotionWidget?.actions = promotion.actions
-                        statusBarPromotionWidget?.notificationType = promotion.notificationType
-
-                        binaryRequestFacade.executeRequest(StatusBarPromotionShownRequest(promotion.message ?: "undefined"))
-                    } else {
-                        clear(statusBarPromotionWidget)
-                    }
-                }
+            timerTask {
+                statusBarUpdater.requestStatusBarMessage()
```
boaz-codota, line 16: NOICE!!!! COOL COOL COOL

src/main/java/com/tabnine/lifecycle/BinaryPromotionStatusBarLifecycle.kt
```diff
@@ -1,61 +1,21 @@
...
 import com.tabnine.general.StaticConfig.BINARY_PROMOTION_POLLING_DELAY
 import com.tabnine.general.StaticConfig.BINARY_PROMOTION_POLLING_INTERVAL
-import com.tabnine.statusBar.StatusBarPromotionWidget
+import com.tabnine.statusBar.StatusBarUpdater
 import java.util.Timer
-import java.util.TimerTask
+import kotlin.concurrent.timerTask
 
 class BinaryPromotionStatusBarLifecycle(private val binaryRequestFacade: BinaryRequestFacade) {
+    private val statusBarUpdater = StatusBarUpdater(binaryRequestFacade)
```
boaz-codota, line 11: Get it in the ctor instead of creating it

src/main/java/com/tabnine/selections/TabNineLookupListener.java
```diff
@@ -22,9 +23,11 @@
 
 public class TabNineLookupListener implements LookupListener {
     private final BinaryRequestFacade binaryRequestFacade;
+    private final StatusBarUpdater statusBarUpdater;
 
     public TabNineLookupListener(BinaryRequestFacade binaryRequestFacade) {
         this.binaryRequestFacade = binaryRequestFacade;
+        this.statusBarUpdater = new StatusBarUpdater(binaryRequestFacade);
```
boaz-codota, line 30: get it in the ctor here as well. Initialize it in the `DependencyContainer` thingy

src/main/java/com/tabnine/statusBar/StatusBarUpdater.kt
```diff
@@ -0,0 +1,94 @@
...
+            }
+
+            val promotion = binaryRequestFacade.executeRequest(StatusBarPromotionBinaryRequest())
+
+            if (promotion != null) {
+                updateStatusBars(statusBarPromotionWidgets, promotion)
+                binaryRequestFacade.executeRequest(
+                    StatusBarPromotionShownRequest(
+                        promotion.id,
+                        promotion.message ?: "undefined", promotion.notificationType,
```
boaz-codota, line 39: extract `"undefined"` to a const with a meaningful name, it looks like a bug here (I know it is not)

src/main/java/com/tabnine/statusBar/StatusBarUpdater.kt
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    private fun clear(statusBarPromotionWidgets: List<StatusBarPromotionWidget.StatusBarPromotionComponent>) {
+        updateStatusBars(
+            statusBarPromotionWidgets, false, null,
+            null, null, null
+        )
+    }
+
+    private fun updateStatusBars(
```
boaz-codota, line 70: This function gets too many arguments, please refactor it so it will get less (create an object that contains many of the arguments this function get). A good object will probably make line 64-67 much more readable (as you can replace the false and null values with a single null value).

src/main/java/com/tabnine/statusBar/StatusBarUpdater.kt
```diff
@@ -0,0 +1,94 @@
...
+            statusBarPromotionWidget.actions = actions
+            statusBarPromotionWidget.notificationType = notificationType
+        }
+    }
+
+    private fun getStatusBarsWidgets(): List<StatusBarPromotionWidget.StatusBarPromotionComponent> {
+        return ProjectManager.getInstance().openProjects.mapNotNull {
+            val statusBar = WindowManager.getInstance()?.getStatusBar(it)
+            val widget = statusBar?.getWidget(StatusBarPromotionWidget::class.java.name) ?: null
+            (widget as StatusBarPromotionWidget).component as StatusBarPromotionWidget.StatusBarPromotionComponent
```
boaz-codota, line 91: Isn't that a possible bug? `widget` could be `null` and when you ask for `.component` it will through `NullPointerException`. Or am I missing something here?

boaz-codota: By the way, Kotlin does not yell on you here because you convert the widget to non nullable. If you'd do `(widget as StatusBarPromotionWidget?)` I think it would not compile

boaz-codota (commented):
I think there is a bug, plus some suggestions.

src/main/java/com/tabnine/lifecycle/BinaryPromotionStatusBarLifecycle.kt
shaiahr: Done (though I think this is over encapsulation).

src/main/java/com/tabnine/statusBar/StatusBarUpdater.kt
shaiahr: Why do we even pass "undefined"? I might set it to 'undefined' here while another plugin will set it to 'empty' etc'.
We should return null and let the binary replace this with whatever 'empty' value it wants.

boaz-codota: 🤷

src/main/java/com/tabnine/statusBar/StatusBarUpdater.kt
shaiahr: You are correct. Wasn't sure about the correct syntax for this. Thanks.

src/main/java/com/tabnine/statusBar/StatusBarUpdater.kt
```diff
@@ -0,0 +1,96 @@
...
+            statusBarPromotionWidget.actions = actions
+            statusBarPromotionWidget.notificationType = notificationType
+        }
+    }
+
+    private fun getStatusBarsWidgets(): List<StatusBarPromotionWidget.StatusBarPromotionComponent> {
+        return ProjectManager.getInstance().openProjects.mapNotNull {
+            val statusBar = WindowManager.getInstance()?.getStatusBar(it)
+            val widget = statusBar?.getWidget(StatusBarPromotionWidget::class.java.name)
+            widget?.let { (widget as StatusBarPromotionWidget).component as StatusBarPromotionWidget.StatusBarPromotionComponent }
```
boaz-codota, line 93: You can write it like so:
```kotlin
            statusBar?.getWidget(StatusBarPromotionWidget::class.java.name)?.let { (it as StatusBarPromotionWidget).component as StatusBarPromotionWidget.StatusBarPromotionComponent }
```

boaz-codota (approved):

PR #122, opened 2021-02-24
Keep status bar message visible for the duration received from the bi…

shaiahr:
…nary.

Remove status bar message when clicked.


boaz-codota (approved):
LGTM

PR #126, opened 2021-03-02
fix broken openhub action + actions type as any

dimacodota:
fix broken openhub action + actions type as any

### DIFF
--- a/src/main/java/com/tabnine/lifecycle/BinaryNotificationsLifecycle.kt
+++ b/src/main/java/com/tabnine/lifecycle/BinaryNotificationsLifecycle.kt
@@ -44,7 +44,7 @@ class BinaryNotificationsLifecycle(
                                             o.actions
                                         )
                                     )
-                                    if (o.actions?.any { it is String && it == OPEN_HUB_ACTION } == true) {
+                                    if (o.actions?.any { it == OPEN_HUB_ACTION } == true) {
                                         actionVisitor.openHub()
                                     }
                                     notification.expire()
--- a/src/main/java/com/tabnine/lifecycle/BinaryNotificationsLifecycle.kt
+++ b/src/main/java/com/tabnine/lifecycle/BinaryNotificationsLifecycle.kt
@@ -44,7 +44,7 @@ class BinaryNotificationsLifecycle(
                                             o.actions
                                         )
                                     )
-                                    if (o.actions?.contains(OPEN_HUB_ACTION) ?: false) {
+                                    if (o.actions?.any { it is String && it == OPEN_HUB_ACTION } == true) {
                                         actionVisitor.openHub()
                                     }
                                     notification.expire()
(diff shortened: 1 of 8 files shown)


src/main/java/com/tabnine/lifecycle/BinaryNotificationsLifecycle.kt
```diff
@@ -44,7 +44,7 @@ class BinaryNotificationsLifecycle(
                                             o.actions
                                         )
                                     )
-                                    if (o.actions?.contains(OPEN_HUB_ACTION) ?: false) {
+                                    if (o.actions?.any { it is String && it == OPEN_HUB_ACTION } == true) {
```
shaiahr, line 47: Why do we need to check the type?

shaiahr: How come '==' is good enough here but in getTooltipText we need equals?

src/main/java/com/tabnine/statusBar/StatusBarPromotionWidget.java
```diff
@@ -89,10 +89,11 @@ public String getTooltipText() {
             if (!e.isPopupTrigger() && MouseEvent.BUTTON1 == e.getButton()) {
                 binaryRequestFacade.executeRequest(new StatusBarPromotionActionRequest(
                         component.getId(), component.getText(), component.getActions()));
-                clearMessage();
-                if (component.getActions().contains(StaticConfig.OPEN_HUB_ACTION)) {
+                List<Object> actions = component.getActions();
+                if(actions != null && actions.stream().anyMatch(a -> a instanceof String && a.toString().equals(StaticConfig.OPEN_HUB_ACTION))){
```
shaiahr, line 93: I think you can just do OPEN_HUB_ACTION.equals(a) and it will check the type (if a is a String, there's no need to call toString on it).

shaiahr (commented):
Just make sure the '==' equality is good enough.

src/main/java/com/tabnine/lifecycle/BinaryNotificationsLifecycle.kt
dimacodota: it is kotlin, the `==` is value equality.
regarding the type check, you are right, ill remove it

src/main/java/com/tabnine/statusBar/StatusBarPromotionWidget.java
```diff
@@ -90,7 +90,7 @@ public String getTooltipText() {
                 binaryRequestFacade.executeRequest(new StatusBarPromotionActionRequest(
                         component.getId(), component.getText(), component.getActions()));
                 List<Object> actions = component.getActions();
-                if(actions != null && actions.stream().anyMatch(a -> a instanceof String && a.toString().equals(StaticConfig.OPEN_HUB_ACTION))){
+                if(actions != null && actions.stream().anyMatch(a -> a.equals(StaticConfig.OPEN_HUB_ACTION))){
```
shaiahr: consider OPEN_HUB_ACTION.equals(a) in case a is null.

shaiahr (approved):
LGTM

PR #141, opened 2021-05-04
sentry added

dimacodota:


src/main/java/com/tabnine/Initializer.java
```diff
@@ -29,6 +31,11 @@ public void beforeApplicationLoaded(@NotNull Application application, @NotNull S
 
     @Override
     public void appStarting(@Nullable Project projectFromCommandLine) {
+        StaticConfig.getSentryDns().ifPresent(sentryDns -> {
+            Sentry.init(options -> {
+                options.setDsn(sentryDns);
+            });
+        });
```
boaz-codota, line 38: I forgot how ugly java is 🤕

src/main/java/com/tabnine/binary/BinaryRequestFacade.java
```diff
@@ -38,16 +39,21 @@ public BinaryRequestFacade(BinaryProcessRequesterProvider binaryProcessRequester
...
             binaryProcessRequesterProvider.onTimeout();
+            Sentry.captureException(e);
         } catch (ExecutionException e) {
             if (e.getCause() instanceof TabNineDeadException) {
                 binaryProcessRequesterProvider.onDead(e.getCause());
             }
 
-            Logger.getInstance(getClass()).warn("Tabnine's threw an unknown error during request.", e);
+            String message = "Tabnine's threw an unknown error during request.";
+            Logger.getInstance(getClass()).warn(message, e);
```
boaz-codota, line 49: I think you can remove the log if we send it to sentry? Is there a way to make sentry write to the logs if he has no connection to the internet?

src/main/java/com/tabnine/general/StaticConfig.java
```diff
@@ -54,12 +57,27 @@
...
+        try (InputStream input = StaticConfig.class.getResourceAsStream("/sentry.properties")) {
+
+            Properties prop = new Properties();
+            if (input != null) {
+                prop.load(input);
+                return Optional.ofNullable(prop.getProperty(SENTRY_DSN));
+            }
+
+        } catch (IOException | RuntimeException e) {
+            // empty, nothing to do here
```
boaz-codota, line 77: maybe add log?

boaz-codota (approved):

src/main/java/com/tabnine/binary/BinaryRequestFacade.java
```diff
@@ -38,16 +39,21 @@ public BinaryRequestFacade(BinaryProcessRequesterProvider binaryProcessRequester
...
+            Sentry.captureException(e, message);
```
amircodota, line 50: I do not like these Sentry calls all over the code.
I think it is better to use the logger, whenever an error/warning is reported, send to sentry too.

They have integration with the populr logging frameworks - here's [an example](https://docs.sentry.io/platforms/java/guides/log4j2/)

If it is a problem to make it work with intellij, we can always create a wrapper around the logger and report to sentry from there.

build.gradle
```diff
@@ -81,6 +82,14 @@ tasks.getByName("runIde").configure {
     }
 }
 
+task createProperty(type: WriteProperties) {
```
amircodota, line 85: Not sure I understand what this means. If I build with a SENTRY_DSN env variable than it will report to sentry, otherwise not?

I think we need to make sure that production builds always have sentry, while other not, at least by default.

And for non production builds, maybe report to a different project or at least tag these so we know they are not production errors

src/main/java/com/tabnine/Initializer.java
```diff
@@ -29,6 +31,11 @@ public void beforeApplicationLoaded(@NotNull Application application, @NotNull S
 
     @Override
     public void appStarting(@Nullable Project projectFromCommandLine) {
+        StaticConfig.getSentryDns().ifPresent(sentryDns -> {
```
amircodota, line 34: What happens if I do not call Sentry.init ? Will calls to sentry error out or will they silently ignore?

amircodota (commented):
see my comments

src/main/java/com/tabnine/Initializer.java
dimacodota: yes

src/main/java/com/tabnine/Initializer.java
```diff
@@ -37,4 +43,16 @@ public void appStarting(@Nullable Project projectFromCommandLine) {
...
+
+    private void initLogger() {
+        Sentry.init();
+        Sentry.configureScope(scope -> {
+            scope.setTag("ide", ApplicationInfo.getInstance().getVersionName());
+            scope.setTag("ideVersion", ApplicationInfo.getInstance().getFullVersion());
+            scope.setTag("os", System.getProperty("os.name"));
+            scope.setTag("channel", Config.CHANNEL);
+        });
+        org.apache.log4j.Logger rootLogger = LogManager.getRootLogger();
```
amircodota, line 55: Hmmm.... won't this send us all sorts of errors from intellij too?
Maybe add it under "com.tabnine" or something like that?

dimacodota: oh, you are right, i added a filter https://github.com/codota/tabnine-intellij/blob/ec999dc55e260d0fa68dfff2cb4487487976b98e/src/main/java/com/tabnine/logging/TabnineFilter.java#L6
do you know a better option for this?

PR #161, opened 2021-06-17
content was updated according to JB request

YotamFromCodota:


dimacodota (approved):
lgtm

src/main/resources/META-INF/plugin.xml
```diff
@@ -252,3 +258,4 @@
         <applicationService serviceImplementation="com.tabnine.lifecycle.BinaryStateService"/>
     </extensions>
 </idea-plugin>
+Tabnine AI code completion for all major programming languages, libraries, and frameworks
```
dimacodota, line 261: why is it here ?

dimacodota (changes requested):
see my comment

PR #178, opened 2021-07-18
Running the analyzing later in write lock

Eransho:
Calling the code analyzer auto import inside a write lock


dimacodota (approved):
lgtm

PR #196, opened 2021-08-19
Feature/apply inline action

Eransho:
Improve the experience and align in JB flow and infrastructure.
Using editor action handler for the accept inline action and implementing our own `EscapeHandler`
Beside that, fixed some bugs like canceling on caret change and not showing the suggestion in the middle of the word


dimacodota (approved):
lgtm

PR #198, opened 2021-08-22
try to recommend tabnine when certain file extensiosn exist

avnerbarr:
Intellij recommend tabnine when:

> The IntelliJ Platform IDEs will recommend a plugin for installation in one of the following cases:
> 
> If a project contains files of types which are unsupported by the running IDE but are supported by a plugin available in the plugin repository.
> 
> If a project was created in n different IDE installation which had the plugin installed, and includes shared settings (modules, facets, artifacts, or run configurations) configured using the plugin.
> 
> 

the instructions are here:
https://plugins.jetbrains.com/docs/marketplace/intellij-plugin-recommendations.html#file-type

(they aren't good)

Tested locally by debugging and witnessing that `.tabnine.model` files have the tabnine icon in the project tree.

> To verify that the file type is registered correctly, you can implement the LanguageFileType.getIcon() method and verify that the correct icon (see Working with Icons and Images) is displayed for files associated with your file type.

We need to release the plugin and test once the static automation tool runs 

https://github.com/JetBrains/intellij-plugin-verifier/tree/master/intellij-feature-extractor/

This is a tool used by the Marketplace to determine a set of supported plugin's features.

> The features are specified using IntelliJ API. In most cases the features are specified as string constants. The tool extracts the constant values from class files using bytecode analysis. There are tricky cases which are hard to analyse without running the plugin's code. In such cases the feature extractor may return incomplete results. The list of extracted plugin features can be seen on this page.

> File extensions
> A plugin author implements abstract class com.intellij.openapi.fileTypes.FileTypeFactory and feeds the com.intellij.openapi.fileTypes.FileTypeConsumer to the only abstract method with supported file extensions.

### DIFF
--- a/src/main/java/com/tabnine/TabnineFileTypeFactory.java
+++ b/src/main/java/com/tabnine/TabnineFileTypeFactory.java
@@ -1,0 +1,23 @@
+package com.tabnine;
+
+import com.intellij.openapi.fileTypes.FileTypeConsumer;
+import com.intellij.openapi.fileTypes.FileTypeFactory;
+import org.jetbrains.annotations.NotNull;
+
+// https://plugins.jetbrains.com/docs/marketplace/intellij-plugin-recommendations.html#file-type
+public class TabnineFileTypeFactory extends FileTypeFactory {
+    @Override
+    public void createFileTypes(@NotNull FileTypeConsumer consumer) {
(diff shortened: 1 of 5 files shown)


dimacodota (approved):
lgtm

PR #206, opened 2021-08-23
Inline tab bug fix

Eransho:
Fixing the major bug where the inline completion mute by mistake the editor Tab action.
The problem was missing `isEnable` implementation.
Also in this PR:
- Changed the `AcceptInlineCompletionHandler` to be a subclass of `EditorWriteActionHandler` (eliminating the need of `runWriteCommandAction` call when applying the selected inline completion)
- Change the inline text hint to be more friendly (in case no shortcut is set to one of the actions)
- Remove old and irrelevant entries from the `plugin.xml`


dimacodota (approved):
lgtm

PR #213, opened 2021-09-01
Delete lib directory

### LINKED ISSUE #202: Why does the lib dir exist in the repo?
This file does not seem to be used, since you're using the gradle buildsystem: [lib/gson-2.8.5.jar](https://github.com/codota/tabnine-intellij/blob/master/lib/gson-2.8.5.jar)

gz#5469

solonovamax:
Remove old leftover files that aren't used.

Fixes #202


dimacodota (approved):
LGTM

PR #244, opened 2021-12-20
Fix/missing state

bilucodota:


boaz-codota (approved):
Seems fine, but I have no idea what this is doing, or what it is supposed to fix. Adding an integration test would be much appreciated.

dimacodota (commented):
lgtm

PR #250, opened 2021-12-28
fix more inline things

yonip23:
## Fixes:
1. When you type `(` the IDE auto-fills `)` - Which causing another invocation of the document listener. 
This means that the first invocation will be debounced and the second one will catch - causing the completion to appear after the auto-filled char (instead of before)
image

To solve this, I'm disabling completions in auto import chars completely. (if you're in some brackets and hit `tab`, IJ catches it before us to move the caret one char forward - that's why im disabling this completely)
2. Wrong condition on caret & key listeners - just the wrong boolean condition...
3. Snippets that start with `\n` were presented in the wrong location


src/main/java/com/tabnine/inline/TabnineDocumentListener.java
```diff
@@ -21,69 +22,87 @@
...
-    isMuted.set(false);
-  }
+        if (startOffset > 0 && startOffset < endOffset) {
+            String textIncludingPreviousChar = document.getText(new TextRange(startOffset - 1, endOffset));
 
-  @Nullable
-  private static Editor getActiveEditor(@NotNull Document document) {
-    if (!ApplicationManager.getApplication().isDispatchThread()) {
-      return null;
+            if (AUTO_FILLING_PAIRS.contains(textIncludingPreviousChar)) {
```
yonip23, line 74: the entire doc was formatted on my ctrl+s, this is the main change...

src/main/java/com/tabnine/inline/TabnineDocumentListener.java
```diff
@@ -21,69 +22,87 @@
...
+import java.util.Arrays;
 import java.util.concurrent.atomic.AtomicBoolean;
 
 public class TabnineDocumentListener implements DocumentListener {
 
-  private final InlineCompletionHandler handler = new InlineCompletionHandler(true);
+    private final InlineCompletionHandler handler = new InlineCompletionHandler(true);
 
-  private static final AtomicBoolean isMuted = new AtomicBoolean(false);
+    private static final java.util.List<String> AUTO_FILLING_PAIRS = Arrays.asList("()", "{}", "[]", "''", "\"\"");
```
dimacodota, line 32: add ticks ("`")

dimacodota (approved):
lgtm

src/main/java/com/tabnine/inline/TabnineDocumentListener.java
```diff
@@ -21,69 +22,87 @@
...
+        if (editor == null || project == null || file == null) {
+            return;
+        }
 
-  public static void mute() {
-    isMuted.set(true);
-  }
+        int startOffset = event.getOffset();
+        int endOffset = event.getOffset() + event.getNewLength();
+        int invocationOffset = endOffset;
```
dimacodota, line 69: extract it to `getinvocationOffset`

bilucodota:
lgtm

PR #254, opened 2022-01-03
More fixes to inline UX

yonip23:
## Fixes
1. Debounce document listener for a short amount of time, to tame calls to the listener.
*Why?*
Mainly for snippets - They appear on empty lines, so if you hit a `\n` and the IDE adds indentation (i.e `\n____`) - It will **sometimes** trigger the listener **twice** - for the newline and for the indentation. This causes snippets to appear with the wrong indentation sometimes, and overall a bad behaior. It's also good for auto-filled chars, like brackets - there's no need to trigger for `(` and `)` separately (not at all actually - this is being handled in a separate condition).
*Double debouncing*
There's another debounce inside `InlineCompletionHandler.java` - So why 2 debouncers?
If you'll have only the outer debouncer you wouldn't issue querties to the binary at all if the call is being debounced, and if you'll have only the inner one you'd issue too many requests.
So the solution is:
- Debounce the outer (listener) function by a short amount of time, to prevent too many calls (like explained above).
- In the inner (handler) function - **always query the binary**, and debounce the update of the UI by a few hundreds of ms (to prevent flickering)
2. Fix auto importer removing the first "import" statement in the file
Apparently, one of the available actions were to "remove unused import" which deleted the import statement entirely - Fixed it by filtering out suggestions to *remove* imports, as its not relevant at all while completing suggestions. The purpose of invoking the auto importer after accepting a suggestion is to import a type that is completed in the suggestion and not imported, not to remove unused imports.

### DIFF
--- a/src/main/java/com/tabnine/inline/InlineCompletionHandler.java
+++ b/src/main/java/com/tabnine/inline/InlineCompletionHandler.java
@@ -37,13 +37,13 @@ import java.util.concurrent.Future;
 public class InlineCompletionHandler implements CodeInsightActionHandler {
     private static final String INLINE_DUMMY_IDENTIFIER = "TabnineInlineDummy";
     private static final Set<Character> CLOSING_CHARACTERS = ContainerUtil.set('\'', '"', '`', ']', '}', ')', '>');
-    public static final int DEBOUNCE_MILLIS = 400;
+    public static final int DEBOUNCE_MILLIS = 350;
 
     private final CompletionFacade completionFacade =
             DependencyContainer.instanceOfCompletionFacade();
     private final MessageBus messageBus = ApplicationManager.getApplication().getMessageBus();
-    private final boolean myForward;
     private Future<?> lastPreviewTask = null;
+    private final boolean myForward;
 
     private static boolean isLocked;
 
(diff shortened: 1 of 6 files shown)


src/main/java/com/tabnine/selections/AutoImporter.java
```diff
@@ -32,197 +31,201 @@
...
+            @NotNull PsiFile file,
+            @NotNull RangeHighlighterEx highlighter,
+            @NotNull String highlightedText) {
+        List<HighlightInfo.IntentionActionDescriptor> availableFixes =
+                ShowIntentionsPass.getAvailableFixes(editor, file, -1, highlighter.getEndOffset());
+        List<HighlightInfo.IntentionActionDescriptor> importActions =
+                availableFixes.stream()
+                        .filter(f -> {
+                            String actionText = f.getAction().getText().toLowerCase();
+                            return actionText.contains("import") && !actionText.contains("remove");
```
yonip23, line 132: the entire file has been formatted, that's why the entire file is being shown at the diff. This line is the only relevant change.

src/main/java/com/tabnine/inline/InlineCompletionHandler.java
```diff
@@ -37,13 +37,13 @@
 public class InlineCompletionHandler implements CodeInsightActionHandler {
     private static final String INLINE_DUMMY_IDENTIFIER = "TabnineInlineDummy";
     private static final Set<Character> CLOSING_CHARACTERS = ContainerUtil.set('\'', '"', '`', ']', '}', ')', '>');
-    public static final int DEBOUNCE_MILLIS = 400;
+    public static final int DEBOUNCE_MILLIS = 350;
```
yonip23, line 40: reduce inner debouncing time to offset this with the outer deobucer ~

PR #255, opened 2022-01-04
handle get text exception

yonip23:
`document.getText` throws `IndexOutOfBoundsException` in some cases and its legit - in these cases the text is obviously not auto-filled, so no special action should be taken


dimacodota (approved):
lgtm

PR #263, opened 2022-01-13
send snippet shown event

yonip23:
### Update

The below issue has been solved by using the new api version 4.0.57

---

Current behavior is that if you received a snippet, lets say:
```javascript
function average(arr) {
  // 'let sum = 0;...'
```
and you continue typing:
```javascript
function average(arr) {
  let // 'sum = 0;...'
```
it'll count as 2 separate `SnippetShown` events.

i think its ok, since (at least in this plugin) currently the snippet disappears and re-appears after some debouncing. 
@dimacodota wdyt?

### DIFF
--- a/src/main/java/com/tabnine/inline/InlineCompletionHandler.java
+++ b/src/main/java/com/tabnine/inline/InlineCompletionHandler.java
@@ -219,7 +219,7 @@ public class InlineCompletionHandler implements CodeInsightActionHandler {
     }
 
     private void afterCompletionShown(TabNineCompletion completion) {
-        if (completion.completionKind == CompletionKind.Snippet) {
+        if (completion.completionKind == CompletionKind.Snippet && !completion.isCached) {
             try {
                 this.binaryRequestFacade.executeRequest(new SnippetShownRequest());
             } catch (RuntimeException e) {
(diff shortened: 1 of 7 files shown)


dimacodota (approved):
lgtm

src/main/java/com/tabnine/intellij/completions/CompletionUtils.java
```diff
@@ -13,65 +13,66 @@
...
+                        result.new_prefix,
+                        result.old_suffix,
+                        result.new_suffix,
+                        index,
+                        newPrefix,
+                        CompletionUtils.getCursorPrefix(document, offset),
+                        CompletionUtils.getCursorSuffix(document, offset),
+                        result.origin,
+                        result.completion_kind,
+                        result.is_cached);
```
yonip23, line 50: this is the only relevant line, the rest is due to the foramtter

src/main/java/com/tabnine/inline/InlineCompletionHandler.java
```diff
@@ -219,7 +219,7 @@ private void retrieveAndShowInlineCompletion(
     }
 
     private void afterCompletionShown(TabNineCompletion completion) {
-        if (completion.completionKind == CompletionKind.Snippet) {
+        if (completion.completionKind == CompletionKind.Snippet && !completion.isCached) {
```
yonip23, line 222: this is the usage of `is_cached`

dimacodota (approved):
lgtm

PR #265, opened 2022-01-18
show suggestions only when a single char has been typed

yonip23:
there was this annoying behavior that when you, for example, `ctrl+v` something into the ide tabnine would hop in and suggest. i disabled it by asserting that the event's new length is exactly 1 - meaning that you typed a single char.


talcodota:
what about empty line completions? still counts as single char when you press ENTER, right?

talcodota (approved):
LGTM

yonip23:
> what about empty line completions? still counts as single char when you press ENTER, right?

no actually, totally forgot about it. good catch!

PR #266, opened 2022-01-20
add color picker settings for inline hint

yonip23:
Added `Tabnine: Settings` section to the settings panel:
image

Which controls the color of the inline hint. Don't need to restart after changing color, and the chosen color gets persisted to the disk.

Settings implementation is taken from the docs here: https://plugins.jetbrains.com/docs/intellij/settings-tutorial.html

### DIFF
--- a/src/main/java/com/tabnine/inline/render/GraphicsUtils.kt
+++ b/src/main/java/com/tabnine/inline/render/GraphicsUtils.kt
@@ -3,10 +3,10 @@ package com.tabnine.inline.render
 import com.intellij.openapi.editor.Editor
 import com.intellij.openapi.editor.colors.EditorFontType
 import com.intellij.ui.JBColor
+import com.tabnine.userSettings.AppSettingsState
 import java.awt.Color
 import java.awt.Font
 import java.awt.font.TextAttribute
-import java.util.HashMap
 import kotlin.math.abs
 import kotlin.math.sqrt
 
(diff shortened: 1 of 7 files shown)


bilucodota (approved):

src/main/java/com/tabnine/userSettings/AppSettingsState.kt
```diff
@@ -0,0 +1,35 @@
...
+
+    override fun loadState(state: AppSettingsState) {
+        XmlSerializerUtil.copyBean(state, this)
+    }
+
+    companion object {
+        @JvmStatic
+        val instance: AppSettingsState
+            get() = ApplicationManager.getApplication().getService(AppSettingsState::class.java)
+    }
```
boaz-codota, line 34: static methods should precede any instance methods

src/main/java/com/tabnine/inline/render/GraphicsUtils.kt
```diff
@@ -21,6 +21,11 @@ object GraphicsUtils {
         return Font(attributes)
     }
 
+    val color: Color
+        get() {
+            return Color(AppSettingsState.instance.color)
+        }
```
boaz-codota, line 27: Not sure I understand why this needs to reside here. I don't like a god util classes, and this seems like one.

boaz-codota (approved):
👏

PR #268, opened 2022-01-20
settings panel: log file path, default color checkbox, disable if inline disabled

yonip23:
- **Use default color:** Overrides user's choice with the default one
- **Log File Path**: equivalent to the `idea.properties` one, and its the preferred one (i.e if both are set, the one from the settings panel is taken) 
image


yonip23:
Disabled color picker if inline is turned off: (you cant change the color value either)
image

dimacodota (approved):
lgtm

bilucodota (approved):
lgtm

PR #275, opened 2022-01-31
making the plugin compatible with 2019.3 - latest eap

yonip23:
1. fixed `Initializer` to use an api that's compatible with 2022.1 (eap)
2. compatibility jobs:
  2.1 on push: check compatibility against 2019.3 and current stable
  2.2 nightly: check compatibility against latest eap

### DIFF
--- a/.github/workflows/compatibility.yml
+++ b/.github/workflows/compatibility.yml
@@ -31,7 +31,7 @@ jobs:
           path: build/reports
 
   compatibility:
-    name: Ensure plugin compatibility against IDEA Ultimate 2018.2, and 2019.3 for IDEA Community, IDEA Ultimate, PyCharm Community, CLion, and the latest EAP snapshot of IDEA Community.
+    name: Ensure plugin compatibility against IC releases 2019.3 - latest eap
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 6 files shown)


build.gradle
```diff
@@ -91,3 +93,17 @@ tasks {
         dependsOn(ktlintFormat)
     }
 }
+
+static def getIdeVersions(Project project) {
+    def ideVersions = project.hasProperty('ideVersions') ? Arrays.asList(project.ideVersions.split(',')) :
```
amircodota, line 98: isn't there a way to say ['2019.3.5', 'latest'] or something like that? no one will remember to update this

amircodota (commented):
see my comment

build.gradle
yonip23: there's no 'latest' option, i tried... but there's no need - every stable version goes through eap beforehand - which means the ci will catch it.
this is how the old gh action meant to work - i think its good enough.
i can make the eap job scheduled for every week or so (in addition to push event), so that even if we don't open a pr for a long time we'd still catch compatibility issues.
wdyt?

.github/workflows/compatibility.yml
```diff
@@ -50,25 +50,35 @@ jobs:
         run: ./gradlew buildPlugin
 
       - name: Validate the plugin
-        run: ./gradlew buildPlugin
+        run: ./gradlew runPluginVerifier
+
+  compatibility-eap:
```
amircodota, line 55: will this run on every PR? it should not fail the PR, only nightly

build.gradle
amircodota: Well, the EAP shoul dnot fail the PR, only notify us so we can prepare.

Maybve you can do something similiar to eap - download latest releast and verify from disk?

yonip23: that's what the eap job does - i extracted it to a nightly job

amircodota (approved):
LGTM

Issue #270, opened 2022-01-21
Frequent IllegalArgumentExceptions thrown on WebStorm 2021.3.1, plugin v0.5.14


ristomatti:
I've recently started getting the below error on WebStorm. It occurs at seemingly random times, every few hours while working on the IDE. I have not yet been able to identify any specific action to reproduce this.I believe this started occurring with the latest plugin OR WebStorm (both were upgraded around the same time).

I'm running WebStorm in Linux Mint 20 (Ubuntu 20.04 LTS) on Intel hardware. The codebase I've recently been working on is mostly TypeScript (Next.js/React).

![image](https://user-images.githubusercontent.com/9029939/150569734-a46a7c10-6fe0-4f63-842f-429a5b5a8611.png)

```
java.lang.IllegalArgumentException: Argument for @NotNull parameter 'file' of com/intellij/openapi/vfs/VfsUtilCore.isAncestor must not be null
	at com.intellij.openapi.vfs.VfsUtilCore.$$$reportNull$$$0(VfsUtilCore.java)
	at com.intellij.openapi.vfs.VfsUtilCore.isAncestor(VfsUtilCore.java)
	at com.intellij.lang.typescript.modules.TypeScriptModuleFileReferenceSet.addGlobalTypesContexts(TypeScriptModuleFileReferenceSet.java:191)
	at com.intellij.lang.typescript.modules.TypeScriptModuleFileReferenceSet.computeDefaultContexts(TypeScriptModuleFileReferenceSet.java:173)
	at com.intellij.psi.impl.source.resolve.reference.impl.providers.FileReferenceSet.getDefaultContexts(FileReferenceSet.java:305)
	at com.intellij.psi.impl.source.resolve.reference.impl.providers.FileReference.getContexts(FileReference.java:97)
	at com.intellij.psi.impl.source.resolve.reference.impl.providers.FileReference.lambda$innerResolve$0(FileReference.java:131)
	at com.intellij.openapi.util.RecursionManager$1.computePreventingRecursion(RecursionManager.java:114)
	at com.intellij.openapi.util.RecursionGuard.doPreventingRecursion(RecursionGuard.java:44)
	at com.intellij.openapi.util.RecursionManager.doPreventingRecursion(RecursionManager.java:68)
	at com.intellij.psi.impl.source.resolve.reference.impl.providers.FileReference.innerResolve(FileReference.java:131)
[...]

PR #278, opened 2022-02-01
dont use deprecated api for computeCurrentPrefix

yonip23:
Trying to fix issue https://github.com/codota/tabnine-intellij/issues/270 -
I still haven't pinpointed the root cause of this exception, but I noticed that the last entry in the stack trace that's from our source code is:
```
...
	at com.intellij.codeInsight.completion.CompletionData.findPrefixStatic(CompletionData.java:161)
	at com.tabnine.inline.InlineCompletionHandler.computeCurrentPrefix(InlineCompletionHandler.java:125)
```
And `findPrefixStatic` is deprecated.
So I replaced it with a more simple way that doesn't need this api. This prefix's only usage is to filter out suggestions that their `new_prefix` doesn't start with said `prefix` - which is basically the last word up to the current `offset`.

### DIFF
--- a/src/main/java/com/tabnine/inline/InlineCompletionHandler.java
+++ b/src/main/java/com/tabnine/inline/InlineCompletionHandler.java
@@ -1,7 +1,6 @@
 package com.tabnine.inline;
 
 import com.intellij.codeInsight.CodeInsightActionHandler;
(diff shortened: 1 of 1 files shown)


yonip23:
closing, working on a new pr: https://github.com/codota/tabnine-intellij/pull/279

PR #279, opened 2022-02-01
use old_prefix instead of deprecated api

yonip23:
Follow up on https://github.com/codota/tabnine-intellij/pull/278

Instead of the deprecated api `CompletionData.findPrefixStatic`, using `old_prefix` field from the binary (This is also how our vscode plugin handles it).

### DIFF
--- a/src/main/java/com/tabnine/inline/InlineCompletionHandler.java
+++ b/src/main/java/com/tabnine/inline/InlineCompletionHandler.java
@@ -1,9 +1,6 @@
 package com.tabnine.inline;
 
 import com.intellij.codeInsight.CodeInsightActionHandler;
-import com.intellij.codeInsight.completion.CompletionData;
-import com.intellij.codeInsight.completion.PlainPrefixMatcher;
-import com.intellij.codeInsight.completion.PrefixMatcher;
 import com.intellij.openapi.application.ApplicationManager;
 import com.intellij.openapi.diagnostic.Logger;
 import com.intellij.openapi.editor.Document;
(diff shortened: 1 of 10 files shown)


src/main/java/com/tabnine/inline/InlineCompletionHandler.java
```diff
@@ -75,14 +68,13 @@ void invoke(@NotNull Project project, @NotNull Editor editor, @NotNull PsiFile f
         final CompletionState completionState = CompletionState.findOrCreateCompletionState(editor);
         int lastDisplayedCompletionIndex = completionState.lastDisplayedCompletionIndex;
 
-        boolean noOldSuggestion = lastDisplayedCompletionIndex == -1 || completionState.prefix == null;
```
dimacodota, line 78: r u sure?i think this can lead to  a  bug,

dimacodota (approved):
see my question

src/main/java/com/tabnine/inline/InlineCompletionHandler.java
yonip23: Afaik the condition was redundant, but just dor the sake of keeping the change as little as possible i can add a check on `oldPrefix`, which is the field that's replacing the `prefix`. Wdyt?

dimacodota: i'm not sure, the `completionState.prefix == null` was like "the first time" check(it would be inited in the if below), checking the prefix would not have the same effect

PR #283, opened 2022-02-14
fix compatibility issues against latest eap

yonip23:
### What is broken
The latest EAP is moved from log4j based logger to `java.util.logging` logger. We are using log4j appender to send error reports to our services.
This means that in the new EAP not only does the appender fails to compile, it wouldn't do anything at all - because no log4j logger is running.

### The fix
There are a few approaches, but I chose to "hack" IJ's plugin SDK logging system and add our logic before each call to the logger.
Because the api is not built for this use case, there's a bit of an acrobatic code there in the package `logging`, but not too much imo.


yonip23:
lateste eap compatibility check on this branch passes: https://github.com/codota/tabnine-intellij/runs/5189478567?check_suite_focus=true

src/main/java/com/tabnine/logging/TabnineLogDispatcher.kt
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    @Throws(Exception::class)
+    fun dispatchLog(level: String, message: String, stackTrace: String?) {
+        ApplicationManager.getApplication()
+            .executeOnPooledThread {
+                val postRequest = HttpPost(String.format("%s/logs/%s", Config.LOGGER_HOST, level))
+                val requestBody = baseRequestBody.deepCopy()
+                requestBody.addProperty("message", message)
+                // requestBody.addProperty("stackTrace", stackTrace);
```
boaz-codota, line 36: no leaving comments behind

boaz-codota (approved):
LGTM

boaz-codota:
BUT: I think you could add testing against the latest heap in the CI @yonip23

yonip23:
> BUT: I think you could add testing against the latest heap in the CI @yonip23 . Or is it already happening at https://github.com/codota/tabnine-intellij/blob/master/.github/workflows/compatibility-eap.yml#L10 ?

The latest eap job runs daily, like the nightly check we've got

yonip23:
> lateste eap compatibility check on this branch passes: https://github.com/codota/tabnine-intellij/runs/5189478567?check_suite_focus=true

This is the job @boaz-codota

src/main/java/com/tabnine/logging/TabnineLogDispatcher.kt
yonip23: @yairco1990 im removing it then

amircodota (approved):
LGTM

yairco1990 (approved):
lgtm

PR #284, opened 2022-02-15
fix kotlin warnings + remove filetype factory

yonip23:
`FileTypeFactory` is deprecated - replaced according to this doc: https://plugins.jetbrains.com/docs/intellij/language-and-filetype.html#register-the-filetype

and it seems to work:
image


avnerbarr (approved):

yairco1990 (approved):
lgtm

PR #289, opened 2022-02-23
remove out of date plugin disable listener

yonip23:
Removing `TabNineDisablePluginListener`, because `PluginManagerCore.addDisablePluginListener` is scheduled for removal in 2021.3, and in the latest release of IJ it started to call this listener on update too - which is false positive.

The new recommendation to add a disable listener is `DisabledPluginsState.addDisablePluginListener`, which is not available for SDK version 2019.3, which is what we're using.

---

This means that when a user unchecks the checkbox here, we won't get an uninstall event anymore.
image


dimacodota (approved):
lgtm

boaz-codota:
> The new recommendation to add a disable listener is DisabledPluginsState.addDisablePluginListener, which is not available for SDK version 2019.3, which is what we're using.

Can't we add this under a flag or something?

PR #290, opened 2022-02-23
promote experimental tabnine inlay to default

yonip23:
`TabnineInlay` is the component that draws the inline hint on the graphics element of the IDE.
We had 2 of those - one was the default, and one available for alpha only.
We have finally decided that the experimental one is a significant enough improvement to be the default one.


dimacodota (approved):
LGTM

PR #301, opened 2022-04-14
change AppSettingsState component name to avoid collisions

yonip23:
the previous name was one taken from a general example somewhere on the internet - so if any other random plugin copied the same example we'll have collisions.
### Note
This is a breaking change, i.e user settings will be reset to default.
We only have the log file path & inline color there, so i dont think its too much trouble, but worth mentioning.


dimacodota (approved):
lgtm

PR #310, opened 2022-04-28
fix snippets tabs

yonip23:
### What?
When showing inline completions, replacing `\t`s with the appropriate amount of spaces.

### Why?
It happens to be the case that `Graphics.drawString` counts `\t` as 0 length - you can also check that `FontMetrics.stringWidth` returns 0 for `\t`.

Before:
image

After:
image


src/main/java/com/tabnine/inline/render/GraphicsUtils.kt
```diff
@@ -54,3 +56,11 @@ object GraphicsUtils {
         )
     }
 }
+
+fun tabSize(editor: Editor): Int {
```
yonip23, line 60: Inspired by some stackoverflow answer

dimacodota (approved):
lgtm

PR #312, opened 2022-05-02
trim snippets with respect to indentation

yonip23:
See the description on pr 1805 in our binary project for more details.

### DIFF
--- a/src/main/java/com/tabnine/binary/requests/autocomplete/CompletionPostprocess.kt
+++ b/src/main/java/com/tabnine/binary/requests/autocomplete/CompletionPostprocess.kt
@@ -24,16 +24,14 @@ fun postprocess(request: AutocompleteRequest, result: AutocompleteResponse, tabS
  * returning `null` if `value` has no newlines, or the last line is not whitespaces only.
  */
 fun lastLineIndentation(value: String, tabSize: Int): Int? {
-    try {
-        val lastLineStartIndex = value.lastIndexOf('\n') + 1
+    val lastLineStartIndex = value.lastIndexOf('\n')
+    if (lastLineStartIndex == -1) return null
 
-        val lastLine = value.substring(lastLineStartIndex).replace("\t", " ".repeat(tabSize))
-        if (lastLine.isBlank()) {
-            return lastLine.length
-        }
-    } catch (e: java.util.NoSuchElementException) {
-        // do nothing
+    val lastLine = value.substring(lastLineStartIndex + 1).replace("\t", " ".repeat(tabSize))
+    if (lastLine.isBlank()) {
+        return lastLine.length
     }
+
     return null
 }
 
--- a/src/main/java/com/tabnine/binary/requests/autocomplete/CompletionPostprocess.kt
+++ b/src/main/java/com/tabnine/binary/requests/autocomplete/CompletionPostprocess.kt
@@ -13,6 +13,10 @@ fun postprocess(request: AutocompleteRequest, result: AutocompleteResponse, tabS
     }
 }
 
+/**
+ * Finds the amount of spaces or tabs in the last line of the given `value`,
+ * returning `null` if `value` has no newlines, or the last line is not whitespaces only.
+ */
 fun lastLineIndentation(value: String): Int? {
     try {
         val lastLine = value.lines().last()
@@ -25,11 +29,18 @@ fun lastLineIndentation(value: String): Int? {
     return null
 }
 
+/**
+ * Constructs a regex which accepts a \n followed by at most `indentation - 1` spaces,
+ * followed by any text or another \n.
+ */
 fun constructRegex(indentation: Int): Regex {
     val upperLimit = indentation - 1
     return Regex("""^ {0,$upperLimit}[\w\n]+""", RegexOption.MULTILINE)
 }
 
+/**
+ * Finds the first match of the given `regex` in `result`, *after* the first newline appearance
+ */
 fun calculateTrimmingIndex(result: String, regex: Regex): Int? {
     val indexOfFirstNewline = result.indexOf("\n")
     if (indexOfFirstNewline < 0) return null
--- a/src/main/java/com/tabnine/binary/requests/autocomplete/CompletionPostprocess.kt
+++ b/src/main/java/com/tabnine/binary/requests/autocomplete/CompletionPostprocess.kt
@@ -1,13 +1,20 @@
 package com.tabnine.binary.requests.autocomplete
 
-fun postprocess(request: AutocompleteRequest, result: AutocompleteResponse, tabSize: Int) {
-    result.results.forEach { it.new_prefix = it.new_prefix.replace("\t", " ".repeat(tabSize)) }
+import com.tabnine.general.CompletionKind
 
-    val requestIndentation = lastLineIndentation(request.before.replace("\t", " ".repeat(tabSize))) ?: return
+fun postprocess(request: AutocompleteRequest, result: AutocompleteResponse, tabSize: Int) {
+    val resultsSubset = result.results.filter { it.completion_kind == CompletionKind.Snippet }
+    if (resultsSubset.isEmpty()) {
+        return
+    }
+
+    resultsSubset.forEach { it.new_prefix = it.new_prefix.replace("\t", " ".repeat(tabSize)) }
+
+    val requestIndentation = lastLineIndentation(request.before, tabSize) ?: return
     if (requestIndentation == 0) return
 
     val regex = constructRegex(requestIndentation)
-    result.results.forEach { entry ->
+    resultsSubset.forEach { entry ->
         val calculateTrimmingIndex = calculateTrimmingIndex(entry.new_prefix, regex)
         calculateTrimmingIndex?.let { entry.new_prefix = entry.new_prefix.take(it) }
     }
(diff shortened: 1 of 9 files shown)


src/main/java/com/tabnine/binary/requests/autocomplete/CompletionPostprocess.kt
```diff
@@ -0,0 +1,48 @@
...
+    }
+}
+
+/**
+ * Finds the amount of spaces or tabs in the last line of the given `value`,
+ * returning `null` if `value` has no newlines, or the last line is not whitespaces only.
+ */
+fun lastLineIndentation(value: String): Int? {
+    try {
+        val lastLine = value.lines().last()
```
boaz-codota, line 22: Is there a way to scan from the end for the 1st linebreak? Won't it be faster for larger texts?

src/main/java/com/tabnine/binary/requests/autocomplete/CompletionPostprocess.kt
```diff
@@ -0,0 +1,48 @@
...
+        if (lastLine.isBlank()) {
+            return lastLine.length
+        }
+    } catch (e: java.util.NoSuchElementException) {
```
boaz-codota, line 26: When is this exception raised? Upon no lines at all?

src/main/java/com/tabnine/binary/requests/autocomplete/CompletionPostprocess.kt
```diff
@@ -0,0 +1,48 @@
+package com.tabnine.binary.requests.autocomplete
+
+fun postprocess(request: AutocompleteRequest, result: AutocompleteResponse, tabSize: Int) {
```
boaz-codota, line 3: I think `tabSize` is constant, and we should calculate the entire regex only once. As far as I know, compiling regexes could have a performance penalty.

boaz-codota: And it is best if we can calculate it on plugin start.

src/main/java/com/tabnine/binary/requests/autocomplete/CompletionPostprocess.kt
```diff
@@ -0,0 +1,48 @@
...
+    result.results.forEach { it.new_prefix = it.new_prefix.replace("\t", " ".repeat(tabSize)) }
+
+    val requestIndentation = lastLineIndentation(request.before.replace("\t", " ".repeat(tabSize))) ?: return
+    if (requestIndentation == 0) return
+
+    val regex = constructRegex(requestIndentation)
+    result.results.forEach { entry ->
+        val calculateTrimmingIndex = calculateTrimmingIndex(entry.new_prefix, regex)
+        calculateTrimmingIndex?.let { entry.new_prefix = entry.new_prefix.take(it) }
```
boaz-codota, line 12:
```suggestion
        calculateTrimmingIndex(entry.new_prefix, regex)?.let { trimmingIndex -> entry.new_prefix = entry.new_prefix.take(trimmingIndex) }
```

src/main/java/com/tabnine/binary/requests/autocomplete/CompletionPostprocess.kt
```diff
@@ -0,0 +1,48 @@
+package com.tabnine.binary.requests.autocomplete
+
+fun postprocess(request: AutocompleteRequest, result: AutocompleteResponse, tabSize: Int) {
+    result.results.forEach { it.new_prefix = it.new_prefix.replace("\t", " ".repeat(tabSize)) }
+
+    val requestIndentation = lastLineIndentation(request.before.replace("\t", " ".repeat(tabSize))) ?: return
+    if (requestIndentation == 0) return
```
boaz-codota, line 7: If you have something like that prefix:
```kotlin
/**
 * Finds the amount of spaces or tabs in the last line of the given `value`,
 * returning `null` if `value` has no newlines, or the last line is not whitespaces only.
 */
fun lastLineIndentation(value: String): Int? {
    try {
        val lastLine = value.lines().last()
```

Where would the following snippet end?
```kotlin
        if (lastLine.isBlank()) {
            return lastLine.length
        }
    } catch (e: java.util.NoSuchElementException) {
        // do nothing
    }
    return null
}

/**
 * Constructs a regex which accepts a \n followed by at most `indentation - 1` spaces,
 * followed by any text or another \n.
 */
fun constructRegex(indentation: Int): Regex {
    val upperLimit = indentation - 1
    return Regex("""^ {0,$upperLimit}[\w\n]+""", RegexOption.MULTILINE)
}
```

src/main/java/com/tabnine/binary/requests/autocomplete/CompletionPostprocess.kt
```diff
@@ -0,0 +1,48 @@
...
+fun constructRegex(indentation: Int): Regex {
+    val upperLimit = indentation - 1
+    return Regex("""^ {0,$upperLimit}[\w\n]+""", RegexOption.MULTILINE)
+}
+
+/**
+ * Finds the first match of the given `regex` in `result`, *after* the first newline appearance
+ */
+fun calculateTrimmingIndex(result: String, regex: Regex): Int? {
+    val indexOfFirstNewline = result.indexOf("\n")
```
boaz-codota, line 45: Send a char and not a string
```suggestion
    val indexOfFirstNewline = result.indexOf('\n')
```

src/main/java/com/tabnine/prediction/CompletionFacade.java
```diff
@@ -74,7 +78,13 @@ private AutocompleteResponse retrieveCompletions(
     req.line = document.getLineNumber(offset);
     req.character = offset - document.getLineStartOffset(req.line);
 
-    return binaryRequestFacade.executeRequest(req, determineTimeoutBy(req.before));
+    AutocompleteResponse autocompleteResponse =
+        binaryRequestFacade.executeRequest(req, determineTimeoutBy(req.before));
+
+    if (autocompleteResponse != null) {
+      postprocess(req, autocompleteResponse, tabSize(editor));
+    }
```
boaz-codota, line 86:
```suggestion
    autocompleteResponse?.let { postprocess(req, it, tabSize(editor)); };
```

src/test/java/com/tabnine/plugin/completionPostProcess/Spaces.kt
```diff
@@ -0,0 +1,51 @@
+package com.tabnine.plugin.completionPostProcess
+
+import com.tabnine.binary.requests.autocomplete.postprocess
+import org.junit.Test
+
+class Spaces {
```
boaz-codota, line 6: kapara give it a bit of a more meaningful name

boaz-codota (approved):
LGTM, but I think we ought to create some performance tests / benchmarks for it. Suggestions are mission critical, and I think we need to act more carefully here. This is another bonus point for Rust as well...

src/main/java/com/tabnine/binary/requests/autocomplete/CompletionPostprocess.kt
yonip23: no, depends on the project / file you're working on - can also be changed by the user..

src/main/java/com/tabnine/binary/requests/autocomplete/CompletionPostprocess.kt
yonip23: its not clear where exactly is the cursor in your example, but assuming you're here:
```kotlin
/**
 * Finds the amount of spaces or tabs in the last line of the given `value`,
 * returning `null` if `value` has no newlines, or the last line is not whitespaces only.
 */
fun lastLineIndentation(value: String): Int? {
    try {
        val lastLine = value.lines().last()
        //<-- cursor here
```

then the snippet will end after the closing bracket of he `if` statement - before the `catch` statement.

src/main/java/com/tabnine/binary/requests/autocomplete/CompletionPostprocess.kt
yonip23: image

yonip23: this is so stupid to throw an exception if the list is empty... why doesn't kotlin return `T?` 😢

src/main/java/com/tabnine/prediction/CompletionFacade.java
yonip23: dis is java here 🤮

src/test/java/com/tabnine/plugin/completionPostProcess/Spaces.kt
yonip23: ah, the test's full name is combined with the package name - which gives it the meaning

src/main/java/com/tabnine/binary/requests/autocomplete/CompletionPostprocess.kt
yonip23: good point - can use somthing like:

```kotlin
value.substring(value.lastIndexOf("\n"))
```

yonip23:
> LGTM, but I think we ought to create some performance tests / benchmarks for it. Suggestions are mission critical, and I think we need to act more carefully here. This is another bonus point for Rust as well...

I get your concern, but I made sure this code will only run from `Snippet` completions - in which the extra few ms are not critical what so ever

PR #313, opened 2022-05-04
prevent tabSize from throwing exception if called without read access

yonip23:


src/main/java/com/tabnine/inline/render/BlockElementRenderer.kt
```diff
@@ -34,12 +34,10 @@ class BlockElementRenderer(
         color = color ?: GraphicsUtils.color
         g.color = color
         g.font = GraphicsUtils.getFont(editor, deprecated)
-        val tabSpaces = " ".repeat(tabSize(editor))
 
         blockText.withIndex().forEach { (i, line) ->
-            val lineWithSpaces = line.replace("\t", tabSpaces)
             g.drawString(
-                lineWithSpaces,
+                line,
```
yonip23, line 40: since the introduction of `postprocess`, this maneuver is redundant

PR #315, opened 2022-05-12
bring back uninstall events

yonip23:
### Context
On February, we disabled the uninstall events because updating the plugin triggers this event as well, and we decided we're better off without this event at all than with ~6x more uninstall events than there should be.

We also tried to use other event listeners to bypass this issue, but it didn't work: https://intellij-support.jetbrains.com/hc/en-us/community/posts/4483992478226-DynamicPluginListener-doesn-t-detect-updates-correctly

Now we decided its a good time to bring them back.
### Solution
We're using the same old listener, and looking at the filesystem to detect whether its an update or an uninstall:

When JetBrains checks for updates (i.e when you open the `Plugins` view), they're downloading the new zip to the path returned from `com.intellij.openapi.applicationPathManager#getPluginTempPath`:

```
❯ ll ~/.cache/JetBrains/IntelliJIdea2022.1/plugins/                                                                                                                                              
total 3.4M
-rw-rw-r-- 1 yoni yoni 1.4M May 12 10:02 brokenPlugins.json
-rw-rw-r-- 1 yoni yoni   36 May 12 10:02 brokenPlugins.json.etag
drwxrwxr-x 2 yoni yoni 4.0K May 11 21:30 meta
-rw-rw-r-- 1 yoni yoni 219K May 12 10:02 pluginsXMLIds.json
-rw-rw-r-- 1 yoni yoni   36 May 12 10:02 pluginsXMLIds.json.etag
-rw-rw-r-- 1 yoni yoni 1.8M May 12 11:42 TabNine.zip  # <----------- This is what we're looking for
```

And this `TabNine.zip` file contains our plugin's jar, as well as some dependencies:
```
❯ unzip -l ~/.cache/JetBrains/IntelliJIdea2022.1/plugins/TabNine.zip                                                                                                                             
Archive:  /home/yoni/.cache/JetBrains/IntelliJIdea2022.1/plugins/TabNine.zip
  Length      Date    Time    Name
---------  ---------- -----   ----
        0  2022-05-12 11:42   TabNine/
        0  2022-05-12 11:41   TabNine/lib/
   193119  2022-05-12 11:42   TabNine/lib/kotlin-stdlib-common-1.4.32.jar
    16221  2022-05-12 11:42   TabNine/lib/kotlin-stdlib-jdk8-1.4.32.jar
   295797  2022-05-12 11:42   TabNine/lib/TabNine-0.6.0.jar # <----------- This is what we're looking for
    22895  2022-05-12 11:42   TabNine/lib/kotlin-stdlib-jdk7-1.4.32.jar
  1495021  2022-05-12 11:42   TabNine/lib/kotlin-stdlib-1.4.32.jar
    17536  2022-05-12 11:42   TabNine/lib/annotations-13.0.jar
     1139  2022-05-12 11:42   TabNine/lib/searchableOptions-0.6.0.jar
---------                     -------
  2041728                     9 files
```

Looking at this jar's version and comparing it to the current running version, we can decide whether the `uninstall` event is false positive or not.

I also verified that this thing works for both IJ2019 and IJ2022 - the value of `getPluginTempPath` changed (i.e different location on the machine), but the whole thing works as intended.


src/main/java/com/tabnine/UninstallListener.kt
```diff
@@ -0,0 +1,62 @@
...
+import com.tabnine.binary.BinaryRequestFacade
+import com.tabnine.binary.requests.uninstall.UninstallRequest
+import com.tabnine.general.StaticConfig
+import com.tabnine.general.readTempTabninePluginZip
+import com.tabnine.lifecycle.UninstallReporter
+import java.nio.file.Paths
+
+private const val TABNINE_JAR_NAME = "TabNine-"
+private const val JAR_SUFFIX = ".jar"
+private val TABNINE_JAR_REGEX = """^$TABNINE_JAR_NAME\d+\.\d+\.\d+.jar""".toRegex()
```
amircodota, line 16: sometimes we release with pre-release - 0.8.0-alpha20220512.....

Probably account for that, ormaybe just check it starts with TabNine- or something

src/main/java/com/tabnine/UninstallListener.kt
```diff
@@ -0,0 +1,62 @@
...
+
+class UninstallListener(private val facade: BinaryRequestFacade, private val uninstallReporter: UninstallReporter) :
+    PluginStateListener {
+
+    override fun install(descriptor: IdeaPluginDescriptor) {
+        // nothing to do here
+    }
+
+    override fun uninstall(descriptor: IdeaPluginDescriptor) {
+        if (descriptor.pluginId?.let { it == StaticConfig.TABNINE_PLUGIN_ID } == false) {
```
amircodota, line 26: try catch on thie method, so that no "error" pops up to the user

amircodota (approved):
LGTM, but see my comments

src/main/java/com/tabnine/UninstallListener.kt
```diff
@@ -0,0 +1,62 @@
...
+            return
+        }
+        if (newerVersionExists(descriptor)) {
+            Logger.getInstance(javaClass)
```
amircodota, line 30: make this logger static?

src/main/java/com/tabnine/UninstallListener.kt
yonip23: i'd rather account for the pre-releases with `...*.jar`, so that i know for sure i can extract the semver from the `d+.d+.d+` section

src/main/java/com/tabnine/UninstallListener.kt
yonip23: wdym, just wrap everything with try catch?

src/main/java/com/tabnine/UninstallListener.kt
yonip23: we use `Logger.getInstance` throughout the entire code, and it returns a static instance anyway

src/main/java/com/tabnine/UninstallListener.kt
amircodota: yes

src/main/java/com/tabnine/UninstallListener.kt
```diff
@@ -10,13 +10,19 @@ import com.tabnine.general.StaticConfig
 import com.tabnine.general.readTempTabninePluginZip
 import com.tabnine.lifecycle.UninstallReporter
 import java.nio.file.Paths
+import java.time.Duration
 
 private const val TABNINE_JAR_NAME = "TabNine-"
 private const val JAR_SUFFIX = ".jar"
-private val TABNINE_JAR_REGEX = """^$TABNINE_JAR_NAME\d+\.\d+\.\d+.jar""".toRegex()
+private val TABNINE_JAR_REGEX = """^$TABNINE_JAR_NAME\d+\.\d+\.\d+(-\w+.\d+)*.jar""".toRegex()
```
yonip23, line 17: catches `-<something>.<number>` 0 or more times, i can make it accept only 0 or 1 times if you think its necessary @amircodota

amircodota, line 17: . -> \.

* -> ?

amircodota (approved):
LGTM

PR #317, opened 2022-05-19
make postprocess work for non-empty lines

yonip23:
### The Bug
Where the last line is empty, the trimming is done correctly:
image

But where you start writing a prefix which matches the snippet, the `postprocess` function is skipping the trimming part:
image

### Solution
Calculate the indentation regardless of any subsequent text in the last line, there's no need to guard this function with the `if (lastLine.isBlank())`.


assapir (approved):

dimacodota (approved):
lgtm

PR #318, opened 2022-05-22
Small refactorings

boaz-codota:
image


src/main/java/com/tabnine/inline/CompletionPreview.java
```diff
@@ -200,14 +168,9 @@ public static CompletionPreview findCompletionPreview(@NotNull Editor editor) {
...
   public static void disposeIfExists(@NotNull Editor editor) {
-    disposeIfExists(editor, preview -> true);
-  }
-
-  static void disposeIfExists(
-      @NotNull Editor editor, @NotNull Predicate<CompletionPreview> condition) {
-    CompletionPreview preview = findCompletionPreview(editor);
-    if (preview != null && condition.test(preview)) {
-      Disposer.dispose(preview);
+    CompletionPreview preview1 = findCompletionPreview(editor);
```
yonip23, line 171: rename `preview1`

src/main/java/com/tabnine/inline/CompletionPreviewInsertionHint.kt
```diff
@@ -0,0 +1,91 @@
...
+                            mouseEvent.source as Component,
+                            point,
+                            editor.component.rootPane.layeredPane
+                        )
+                    )
+                },
+                HINT_DELAY_MS
+            )
+        }
+    }
```
yonip23, line 55: invert condition and early-return to avoid unnecessary indentation:

```suggestion
        if (!isOverPreview(point)) return

        alarm.addRequest(
            {
                InlineHints.showPreInsertionHint(
                    editor,
                    SwingUtilities.convertPoint(
                        mouseEvent.source as Component,
                        point,
                        editor.component.rootPane.layeredPane
                    )
                )
            },
            HINT_DELAY_MS
        )
    }
```

src/main/java/com/tabnine/inline/CompletionPreviewInsertionHint.kt
```diff
@@ -0,0 +1,91 @@
...
+        }
+
+        val pos: LogicalPosition = editor.xyToLogicalPosition(p)
+
+        if (pos.line >= editor.getDocument().getLineCount()) {
+            return false
+        }
+
+        val pointOffset: Int = editor.logicalPositionToOffset(pos)
+        val inlayOffset: Int = inlay.offset ?: return false
```
yonip23, line 87: remove type hint (`: Int`), i think its redundant here

src/main/java/com/tabnine/inline/CompletionPreviewInsertionHint.kt
```diff
@@ -0,0 +1,91 @@
...
+
+        return inlayOffset <= pointOffset && pointOffset <= inlayOffset + suffix.length
```
yonip23, line 89: i think it'll be clearer to rearrange the conditions like so:

```suggestion
        return pointOffset >= inlayOffset && pointOffset <= inlayOffset + suffix.length
```

(x >= lower bound && x <= upper bound)

src/main/java/com/tabnine/inline/TabnineDocumentListener.java
```diff
@@ -26,7 +26,7 @@
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
-public class TabnineDocumentListener implements DocumentListener {
+public class TabnineDocumentListener implements BulkAwareDocumentListener {
```
yonip23, line 29: why changing to `BulkAware`? do we know when exactly does user's changes are applied in `bulk` mode? I think its not all the time, so we'll probably miss on many keystrokes.

yonip23 (commented):
except for the `BulkAwareDocumentListener`, this PR is great.
So im not approving yet, please reply / make changes on this topic.

src/test/java/com/tabnine/plugin/HintTests.kt
```diff
@@ -1,89 +0,0 @@
-package com.tabnine.plugin
```
yonip23, line 1: y are these tests removed?

src/main/java/com/tabnine/inline/TabnineDocumentListener.java
boaz-codota: image
...

src/test/java/com/tabnine/plugin/HintTests.kt
boaz-codota: Because I can't reproduce the settings now that I refactored it. If you know how to solve that, let me know

src/main/java/com/tabnine/inline/CompletionPreviewInsertionHint.kt
boaz-codota: I think you are wrongg, but I changed it

src/main/java/com/tabnine/inline/TabnineDocumentListener.java
yonip23: Lets try this in alpha then, interesting

dimacodota (approved):
lgtm

PR #319, opened 2022-05-26
uninstall fix - stop time check on the newer version zip

yonip23:
Since the latest alpha release caused FP uninstall events internally (@boaz-codota and @amircodota) - 
And since this code has been manually and automatically tested pretty thoroughly -
My current suspicion is that the "stale" check I added causes the FP.
IJ downloads the new plugin version when it decides to check for updates, which is not necessarily related to when the users wants to update.

Note that without the "stale" check, there could be TN now - users who had the new plugin version downloaded and decided to uninstall will not fire an uninstall event.

### DIFF
--- a/src/main/java/com/tabnine/UninstallListener.kt
+++ b/src/main/java/com/tabnine/UninstallListener.kt
@@ -7,6 +7,7 @@ import com.intellij.util.text.SemVer
 import com.tabnine.binary.BinaryRequestFacade
 import com.tabnine.binary.requests.uninstall.UninstallRequest
 import com.tabnine.general.StaticConfig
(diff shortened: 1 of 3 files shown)


dimacodota (approved):
lgtm

amircodota:
@yonip23 Do you see it in the code - that IJ downloads the zip regardless of user update?

yonip23:
@amircodota indeed, when I ran the manual checks I clearly saw that the plugin is being downloaded **at least** whenever you open the `Plugins` view in the settings.
There could be more events when it decides to update, but that's one of them where the user did not click "update" what so ever.

PR #322, opened 2022-05-30
fix uninstall again

yonip23:
Since checking on the zip file didn't work (it appears that JB is downloading the new version only AFTER calling the previous uninstall) --> The patch for now will be to distinguish between uninstall and update according to the stack trace.
See: https://intellij-support.jetbrains.com/hc/en-us/community/posts/4483992478226/comments/5827377386770
image

---

I also added a GH action (already in master) that makes sure the function exists in this class: https://github.com/codota/tabnine-intellij/blob/master/.github/workflows/uninstallIntegrityCheck.yml


src/main/java/com/tabnine/UninstallListener.kt
```diff
@@ -3,27 +3,20 @@ package com.tabnine
...
 import com.tabnine.lifecycle.UninstallReporter
-import java.nio.file.Paths
-import java.time.Duration
 
-private const val TABNINE_JAR_NAME = "TabNine-"
-private const val JAR_SUFFIX = ".jar"
-private val TABNINE_JAR_REGEX = """^$TABNINE_JAR_NAME\d+\.\d+\.\d+(-\w+\.\d+)?.jar""".toRegex()
+// An ugly workaround to distinguish between update and uninstall -->
+// If it's an update, this method name will appear in the stack trace.
+private const val updateMethodName = "installOrUpdatePlugin"
```
dimacodota, line 13: maybe it's safer to check uninstallAndUpdateUi?

yonip23: why would it be better? im ok with either way

dimacodota: because if the  function does not exists(moved/modified) you will not  trigger the uninstall flow

yonip23: oh nice one, will do

yonip23:
i ran manual tests against IJ 2019 and 2022 and its working (removed `last_uninstall`) between each test

dimacodota (approved):
LGTM

PR #328, opened 2022-06-02
Feature/refresh capabilities

amircodota:


src/main/java/com/tabnine/capabilities/CapabilitiesService.java
```diff
@@ -28,27 +28,71 @@ public void init() {
...
+      refreshLoop = new Thread(() -> this.fetchCapabilitiesLoop());
+      refreshLoop.setDaemon(true);
+      refreshLoop.start();
+    }
+  }
+
+  public static final int INITIAL_DELAY_MS = 2000;
+  public static final int LOOP_INTERVAL_MS = 1000;
+
+  public static final int REFRESH_EVERY_MS = 10 * 60 * 1000; // 10 minutes
```
yonip23, line 47: move static ints up

src/main/java/com/tabnine/capabilities/CapabilitiesService.java
```diff
@@ -28,27 +28,71 @@ public void init() {
...
+
+  private void fetchCapabilitiesLoop() {
+    Optional<Long> lastRefresh = Optional.empty();
+    Optional<Long> lastPid = Optional.empty();
+
+    try {
```
yonip23, line 53: why 2 `try` scopes with the same `catch` implementation? imo only the inner one should exist, so that we'll recover from a single loop iteration's failure

src/main/java/com/tabnine/capabilities/CapabilitiesService.java
```diff
@@ -28,27 +28,71 @@ public void init() {
...
+      Thread.sleep(INITIAL_DELAY_MS);
+
+      while (true) {
+        try {
+          Long pid = binaryRequestFacade.pid();
+          if (!lastRefresh.isPresent()
+              || System.currentTimeMillis() - lastRefresh.get() >= REFRESH_EVERY_MS
+              || !lastPid.isPresent()
+              || lastPid.get() == null
+              || !lastPid.get().equals(pid)) {
```
yonip23, line 63: extract to function / variable

src/main/java/com/tabnine/capabilities/CapabilitiesService.java
```diff
@@ -28,27 +28,71 @@ public void init() {
...
+    }
   }
 
   private void fetchCapabilities() {
     final CapabilitiesResponse capabilitiesResponse =
         binaryRequestFacade.executeRequest(new CapabilitiesRequest());
     if (capabilitiesResponse != null && capabilitiesResponse.getEnabledFeatures() != null) {
+      setCapabilities(capabilitiesResponse);
+    } else {
+      scheduleFetchCapabilitiesTask();
```
yonip23, line 86: `scheduleFetchCapabilitiesTask` calls `fetchCapabilities`, which then calls `scheduleFetchCapabilitiesTask` in case of failure - why is this kind of recursion needed? 
once `init` is called, its calling `scheduleFetchCapabilitiesTask` which runs in an infinite loop - so why need to re-schedule in case of failure here?

src/main/java/com/tabnine/capabilities/CapabilitiesService.java
```diff
@@ -28,27 +28,71 @@ public void init() {
...
-    return enabledCapabilities.contains(capability);
   }
 
-  private void scheduleFetchCapabilitiesTask() {
-    fetchCapabilitiesFuture =
-        AppExecutorUtil.getAppScheduledExecutorService()
-            .schedule(this::fetchCapabilities, 2, TimeUnit.SECONDS);
+  private synchronized void scheduleFetchCapabilitiesTask() {
+    if (refreshLoop == null) {
+      refreshLoop = new Thread(() -> this.fetchCapabilitiesLoop());
```
yonip23, line 38:
```suggestion
      refreshLoop = new Thread(this::fetchCapabilitiesLoop);
```

src/main/java/com/tabnine/capabilities/CapabilitiesService.java
yonip23: in addition:
since you're calling `scheduleFetchCapabilitiesTask` from within the thread, which can, in its turn, spawn another one recursively - this function should also be synchronized on `refreshLoop`
(since it can be called from many threads)

yonip23 (changes requested):
overall great work, i like the synchornization on `enabledCapabilities`.
please lmk wdyt about my comments, but i'd like to see the implementation a bit different

src/main/java/com/tabnine/capabilities/CapabilitiesService.java
amircodota: The outer ones is to prevent an exception from reaching the user. For every thread we should probably do this.

The inner one is to continue the loop even of something failed.

I agree that the outer one is very unlikely to happen, but it could happen for example if `Thread.sleep` throws ane wexception.

yonip23 (approved):
🦈

PR #339, opened 2022-06-28
sunset CompletionPostprocess - moved to binary

yonip23:
# DO NOT MERGE UNTIL BINARY 4.4.57 IS IN GA

Sunsetting this feature includes adding `indentation_size` field to `AutocompleteRequest`.


yonip23:
@amircodota this bug caused the function to never read the tab size in production, and always return 4:
image

Now that I fixed it, I see that it wasn't being called from a thread with read access - so I made the function read the tabSize asynchronously, and return the default tabSize only as a fallback until its done

src/main/java/com/tabnine/inline/render/GraphicsUtils.kt
```diff
@@ -15,6 +17,43 @@ import kotlin.math.abs
...
+
+        // Some tests don't run with read access -> can't access tabSize information
+        if (ApplicationManager.getApplication().isUnitTestMode) {
+            return DEFAULT_TAB_SIZE
+        }
+
+        AppExecutorUtil.getAppExecutorService().submit {
+            ApplicationManager.getApplication().runReadAction {
+                Logger.getInstance("GraphicsUtils").info("Reading tab size from editor")
+                tabSize = readTabSize(editor)
```
amircodota, line 35: Hmnm.... I think tabSize can differ in IJ per language. That's why you pass in the editor.

Not sure it will work correctly to store a global tab size

yonip23: i think you're right 🙂 
i can always issue a read operation and return the previous value, wdyt about that?
also, before the bugfix it was always returning 4 and i didn't hear many complains about it - i guess 99% of users have tabSize = 4 anyways...

amircodota: Previous value might still be a different editor, so it is istill not 100% correct.

Can't we request read access? Or maybe call this function fro a different place that has read access?

yonip23:
I verified with debug logs of `ApplicationManager.getApplication().isReadAccessAllowed()` is true in the locations where I added the call to `getTabSize`.

Also verified that if I change the settings to 2, it sends 2 to the binary.

src/main/java/com/tabnine/inline/render/GraphicsUtils.kt
```diff
@@ -59,20 +60,14 @@ object GraphicsUtils {
     }
 }
 
-fun tabSize(editor: Editor): Int? {
-    // Some tests don't run with read access -> can't access tabSize information
-    if (!ApplicationManager.getApplication().isUnitTestMode) {
-        return 4
+fun getTabSize(editor: Editor): Int {
+    if (!ApplicationManager.getApplication().isReadAccessAllowed) {
```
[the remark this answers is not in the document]
yonip23: i can raise an exception, and then the caller can take the default value himself. wdyt?

yonip23:
i made sure that the exception is thrown only for alpha users, and that it actually prevents completions from being shown

amircodota (approved):
LGTM

PR #341, opened 2022-07-03
support hover element for inline completions

yairco1990:


src/main/java/com/tabnine/hover/HoverUpdater.java
```diff
@@ -0,0 +1,163 @@
...
+
+  public void update(final Editor editor) {
+    final int caretOffset = editor.getCaretModel().getOffset();
+    final AtomicReference<Inlay> inlayHolder = new AtomicReference<>();
+    final AtomicBoolean documentChanged = new AtomicBoolean(false);
+    addInlayDisposer(editor, inlayHolder, documentChanged);
+    tryAddLimitExceededInlay(editor, caretOffset, inlayHolder, documentChanged);
+  }
+
+  private void tryAddLimitExceededInlay(
```
yonip23, line 35: rename, `limitExceeded` is specific to the old class

PR #350, opened 2022-07-20
feat: show completion hint on the first suggestion

yairco1990:
https://user-images.githubusercontent.com/10290661/182035131-fe5b5afc-ca7e-4696-bdbe-04e47ac4fd12.mp4


src/main/java/com/tabnine/Initializer.java
```diff
@@ -43,6 +44,7 @@ private void initialize() {
       CapabilitiesService.getInstance().init();
       TabnineUpdater.pollUpdates();
       PluginInstaller.addStateListener(instanceOfUninstallListener());
+      UserState.init();
```
yonip23, line 47: weren't we using `StateRequest` anywhere else? how aobut migrating existing places to use this singleton state?

src/main/java/com/tabnine/balloon/GotItTooltipActions.java
```diff
@@ -0,0 +1,5 @@
+package com.tabnine.balloon;
+
+public interface GotItTooltipActions {
```
yonip23, line 3: its a single action, so `GotItTooltipAction`

yonip23 (changes requested):
Overall good job, the UI looks quite good as well 🙂 

Other than my specific comments, there's something that doesn't sit well with me regarding the lifecycle of ths "got it" ui element - 
It is being disposed and re-rendered on every document change.
I would have expect to see something that is being rendered once upon first hint, and disposed once when we decide to dispose it.

It'll solve 2 issues that I see with the current implementation:
1. The message is "moving around" with the user when they type
2. The message is "blinking" - disappears and re-appears on every key stroke.

I'd also like to see some tests 🙂

yairco1990:
This is how it looks now:
https://user-images.githubusercontent.com/10290661/182035062-7fb764c9-8769-4bba-aa18-04ccff72bea7.mp4

src/main/java/com/tabnine/Initializer.java
yairco1990: I'm not sure that a stale state fits for the other usages. for me it's ok as I'm using only the installation time.

src/main/java/com/tabnine/balloon/GotItTooltip.java
```diff
@@ -0,0 +1,116 @@
...
+import com.intellij.openapi.util.Disposer;
+import com.intellij.ui.JBColor;
+import com.intellij.ui.awt.RelativePoint;
+import com.intellij.util.ui.JBUI;
+import com.tabnine.binary.BinaryRequestFacade;
+import com.tabnine.binary.requests.notifications.shown.HintShownRequest;
+import javax.swing.*;
+import org.jetbrains.annotations.NotNull;
+
+public class GotItTooltip implements Disposable {
```
yonip23, line 24: rename class to `SuggestionHintTooltipUI` or something

src/main/java/com/tabnine/balloon/CompletionHintTooltip.java
```diff
@@ -0,0 +1,51 @@
...
+
+import com.intellij.openapi.diagnostic.Logger;
+import com.intellij.openapi.editor.Editor;
+import com.tabnine.selections.CompletionListener;
+import com.tabnine.selections.CompletionObserver;
+import com.tabnine.state.UserState;
+import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
+
+public class CompletionHintTooltip {
```
yonip23, line 11: rename to `SuggestionHintTooltip`

src/main/java/com/tabnine/balloon/CompletionHintTooltip.java
```diff
@@ -0,0 +1,51 @@
...
+          "first-completion-hint",
+          "Your first completion from Tabnine",
+          "Press <b>Tab</b> to accept Tabnine's suggestion",
+          () -> UserState.getInstance().getCompletionHintState().setIsShown(true));
+
+  public static void handle(Editor editor) {
+    try {
+      UserState userState = UserState.getInstance();
+      if (userState.getCompletionHintState().isEligibleForCompletionHint()) {
+        if (!completionHintTooltip.isVisible()) {
```
yonip23, line 23: invert conditions to something like
```java
if (!userState.getCompletionHintState().isEligibleForCompletionHint() || 
    completionHintTooltip.isVisible()) {
    // "else" code here
    return;
}
// main code here
```

src/main/java/com/tabnine/balloon/CompletionHintTooltip.java
```diff
@@ -0,0 +1,51 @@
...
+              .schedule(
+                  () -> {
+                    UserState.getInstance().getCompletionHintState().setIsShown(true);
+                    completionHintTooltip.dispose();
+                  },
+                  30,
+                  TimeUnit.SECONDS);
+        }
+      } else {
+        userState.getCompletionHintState().setIsShown(true);
```
yonip23, line 44: not sure its important, but why not checking if its already true before setting? `setIsShown` is saving a persistent state so I guess is quite expensive operation

src/main/java/com/tabnine/balloon/GotItTooltip.java
```diff
@@ -0,0 +1,116 @@
...
+
+    panel.add(Box.createVerticalStrut(15));
+    panel.add(gotItButton);
+
+    return panel;
+  }
+
+  private void showTooltip(Editor editor, Balloon tooltip) {
+    RelativePoint relativePoint = JBPopupFactory.getInstance().guessBestPopupLocation(editor);
+    relativePoint.getPoint().translate(-5, -editor.getLineHeight() / 2);
```
yonip23, line 92: extract -5 and 2 to consts

src/main/java/com/tabnine/balloon/GotItTooltipAction.java
```diff
@@ -0,0 +1,5 @@
+package com.tabnine.balloon;
+
+public interface GotItTooltipAction {
+  void onGotItClicked();
+}
```
yonip23, line 5: rename to `SuggestionHintTooltipAction`.
the function as well - `onSuggestionHintTooltipClicked`

src/main/java/com/tabnine/capabilities/Capability.java
```diff
@@ -7,4 +7,7 @@ public enum Capability {
   INLINE_SUGGESTIONS,
   @SerializedName("alpha")
   ALPHA,
+
+  @SerializedName("completion_hint_enabled")
+  COMPLETION_HINT_ENABLED,
```
yonip23, line 12: `SUGGESTION_HINT_ENABLED` - also the value in the binary should be changed

src/main/java/com/tabnine/general/Utils.java
```diff
@@ -70,4 +72,21 @@ public static List<String> asLines(String block) {
...
+
+  public static String wrapWithHtmlTag(String content, String tag) {
+    return "<" + tag + ">" + content + "</" + tag + ">";
+  }
+
+  public static long getDaysDiff(Date date1, Date date2) {
+    try {
+      return TimeUnit.DAYS.convert(
+          Math.abs(date2.getTime() - date1.getTime()), TimeUnit.MILLISECONDS);
+    } catch (NullPointerException e) {
```
yonip23, line 88: why catching nulls here? you can annotate the dates with `@NotNull`

src/main/java/com/tabnine/balloon/GotItTooltip.java
yairco1990: the purpose of this class is to provide other hints we may use in the future, and not just the completion hint

src/main/java/com/tabnine/balloon/CompletionHintTooltip.java
yairco1990: in this specific example, I want to handle the else case of the isEligible condition, and ignore the else case of the isVisible

yairco1990: something like
`      if (!userState.getCompletionHintState().isEligibleForCompletionHint()) {
        userState.getCompletionHintState().setIsShown(true);
        return;
      }
      if (!completionHintTooltip.isVisible()) {
        return;
      }`

src/main/java/com/tabnine/balloon/CompletionHintTooltip.java
yairco1990: you're right. I'll add the checking

src/main/java/com/tabnine/capabilities/Capability.java
```diff
@@ -7,4 +7,7 @@ public enum Capability {
   INLINE_SUGGESTIONS,
   @SerializedName("alpha")
   ALPHA,
+
+  @SerializedName("suggestion_hint_enabled")
```
yonip23, line 11: first_suggestion_hint?

yonip23 (approved):
🚀

src/main/java/com/tabnine/capabilities/Capability.java
yonip23: im worried that we'll have a feature which includes suggestion hints, but unrelated to this feature at all - then we'll have a bit of a hard time with the naming convention. 
i'll leave it to your decision, not critical 🙂

yairco1990: yeah, that could happen. yalla, I'm changing it. :)

PR #356, opened 2022-08-03
fix snippet shown event

yonip23:
the context field was being sent with camelCase, while the binary expected snake_case.
this, combined with the relevant binary PR created the situation where no snippet related fields where added to the event


dimacodota (approved):
lgtm

PR #365, opened 2022-08-14
hybrid inline/popup mode

yonip23:
Snippet completions are shown in inline, and all others are shown in popup.
This behavior depends on a capability called `"use_hybrid_inline_popup"` - which the binary returns true only for alpha users currently.


yonip23:
In order to write the hybrid mode tests i needed to mock `SuggestionsMode::getSuggestionsMode` - which works fine for the inline flow, but no matter what I tried, the popup completions contributor (`TabNineCompletionContributor`) did not respect my static mock, and executed the actual method instead.
I tried to staticly mock a different class there just to check, and it didn't respect this mock as well...

I ended up creating a `SuggestionsModeService`, which is being created at the `DependencyContainer` and has a **non static** method `getSuggestionsMode`.
I changed all usages to have a data member of this service and call it instead of the static method, and now the popup contributor respects this mock as well.

src/main/java/com/tabnine/inline/InlineCompletionHandler.java
```diff
@@ -74,6 +82,9 @@ private void rerenderCompletion(
       List<TabNineCompletion> completions,
       int offset,
       long modificationStamp) {
+    if (suggestionsModeService.getSuggestionMode() == SuggestionsMode.HYBRID) {
+      completions.removeIf(completion -> !completion.isSnippet());
+    }
```
yonip23, line 87: usage of the hybrid mode

src/main/java/com/tabnine/intellij/completions/TabNineCompletionContributor.java
```diff
@@ -53,6 +56,12 @@ public void fillCompletionVariants(
     if (originalMatcher.getPrefix().length() == 0 && completions.results.length == 0) {
       return;
     }
+
+    if (suggestionsModeService.getSuggestionMode() == SuggestionsMode.HYBRID
+        && Arrays.stream(completions.results).anyMatch(Completion::isSnippet)) {
+      return;
+    }
```
yonip23, line 63: usage of the hybrid mode - dont render popup completions if they contain snippets

yairco1990 (approved):
pagaz

src/main/java/com/tabnine/intellij/completions/TabNineCompletionContributor.java
yairco1990: why not remove the snippet completions as you did in the InlineCompletionHandler?

yonip23: Good catch 🙂
I saw it on dima's branch in vscode and thought so as well, but there's a reason for doing so - 

When you get a snippet and type the same prefix as the snippet, you'll get both inline and popup completions, and we don't want that (at least currently)

yairco1990: got it! thanks :)

src/main/java/com/tabnine/inline/TabnineDocumentListener.java
```diff
@@ -59,7 +61,7 @@ public void documentChanged(@NotNull DocumentEvent event) {
   private boolean shouldIgnoreChange(DocumentEvent event, Editor editor, int offset) {
     Document document = event.getDocument();
 
-    if (SuggestionsMode.getSuggestionMode() != SuggestionsMode.INLINE || event.getNewLength() < 1) {
+    if (event.getNewLength() < 1 || !suggestionsModeService.getSuggestionMode().isInlineEnabled()) {
```
yonip23, line 64: switched conditions here because for some reason the tests in `PredictionBehaviourIntegrationTests.java` yielded a doc change with empty string (they shouldn't 🤷🏻), and `suggestionsModeService.getSuggestionMode()` returned `null` 😢 
Looks like my kombina with this service helped the other tests and fucked up these tests... I really can't figure out why, both the hybrid tests and the tests in `PredictionBehaviourIntegrationTests` are extending the same abstract class with the same mocks and flow, I can't figure out what's different :(

Well, as long as it works...
@dimaabramovich @yairco1990

yonip23: Found the issue, the difference was that the tests in `PredictionBehaviourIntegrationTests` didn't register the immediate application:
```java
    ApplicationManager.setApplication(mockedApplicationWhichInvokesImmediately(), this);
```
I moved this logic to the abstract class since there's no reason not to use it in all these tests 🙂

yairco1990 (approved):
lgtm again

PR #373, opened 2022-09-01
change rendering mode reporting

yonip23:
Until now it reported the rendering mode determined by the binary, but this is not what we're looking for with this report - we want to know per selection how it was **actually** rendered - so sending "hybrid" doesn't say anything.

I created another enum which represents the actual rendering that took affect, and separated it from the "suggestions mode" which the binary determines.


yairco1990 (approved):
shkoyeh

PR #375, opened 2022-09-01
feature: render inline suggestions with the same prefix as the selected lookup item

yairco1990:
https://user-images.githubusercontent.com/10290661/187948315-d5384b24-b771-4b67-b05c-0f717ee1de9b.mp4


src/main/java/com/tabnine/inline/CompletionAdjustment.java
```diff
@@ -0,0 +1,12 @@
+package com.tabnine.inline;
+
+import com.tabnine.binary.requests.autocomplete.AutocompleteRequest;
+import com.tabnine.binary.requests.autocomplete.AutocompleteResponse;
+
+public interface CompletionAdjustment {
+  void adjustRequest(AutocompleteRequest autocompleteRequest);
```
yonip23, line 7: mmm i prefer the signature to be `AutocompleteRequest adjustRequest(AutocompleteRequest autocompleteRequest);` - such that the function takes an object and returns a processed object, instead of mutating it internally. its less side-effect-y this way

src/main/java/com/tabnine/inline/CompletionAdjustmentType.java
```diff
@@ -0,0 +1,5 @@
+package com.tabnine.inline;
+
+public enum CompletionAdjustmentType {
+  LookAhead
```
yonip23, line 4: what's the value of enum with one variant?

yonip23 (approved):
great work overall, 2 things:
1. please try to re-write the new files in kotlin
2. lets report this in the suggestion / selection requests as well - this will require a binary version as well, so i dont mind merging this pr without the analytics so we can try it intenrally.
but i do think we'll want a way to report lookahead suggestions and completions to the binary. - consult with @dimaabramovich maybe?

yairco1990:
> great work overall, 2 things:
> 
> 1. please try to re-write the new files in kotlin
> 2. lets report this in the suggestion / selection requests as well - this will require a binary version as well, so i dont mind merging this pr without the analytics so we can try it intenrally.
>    but i do think we'll want a way to report lookahead suggestions and completions to the binary. - consult with @dimaabramovich maybe?

Thank's bro!
1. sure, I'll do that.
2. why do we need a new binary version for this?

yonip23:
> > great work overall, 2 things:
> > 
> > 1. please try to re-write the new files in kotlin
> > 2. lets report this in the suggestion / selection requests as well - this will require a binary version as well, so i dont mind merging this pr without the analytics so we can try it intenrally.
> >    but i do think we'll want a way to report lookahead suggestions and completions to the binary. - consult with @dimaabramovich maybe?
> 
> Thank's bro!
> 1. sure, I'll do that.
> 2. why do we need a new binary version for this?

Because you'll send an additional field on the suggestion and the selection requests

src/main/java/com/tabnine/inline/CompletionAdjustmentType.java
yairco1990: because of the condition where I check if the snippets should be removed.
I can also add a function that asks if snippets should be removed inside the interface, but I'm not sure it's related to the condition's context. do you have another idea?

PR #376, opened 2022-09-08
feat: adding debounce between completions

yairco1990:


src/main/java/com/tabnine/inline/CompletionTracker.kt
```diff
@@ -0,0 +1,24 @@
+package com.tabnine.inline
+
+import com.intellij.openapi.editor.Editor
+import com.intellij.openapi.util.Key
+import com.tabnine.userSettings.AppSettingsState.Companion.instance
+
+object CompletionTracker {
+    private val LAST_COMPLETION_REQUEST_TIME = Key.create<Long>("LAST_COMPLETION_REQUEST_TIME")
```
yonip23, line 8: who's propagating values into this?

src/main/java/com/tabnine/userSettings/AppSettingsComponent.kt
```diff
@@ -19,6 +19,7 @@ class AppSettingsComponent {
     private val suggestionsModeService = DependencyContainer.instanceOfSuggestionsModeService()
     private val logFilePathComponent = JBTextField()
     private val logLevelComponent = JBTextField()
+    private val debounceTimeComponent = JBTextField()
```
yonip23, line 22: are u sure you want to expose this? i think we need to come up with the correct value ourselves, not sure it should be in the user settings panel.
if you want to easily configure it, put it in a secret file or env var, no?

src/main/java/com/tabnine/inline/InlineCompletionHandler.java
```diff
@@ -65,22 +69,24 @@ public void retrieveAndShowCompletion(
       }
     } else {
       ObjectUtils.doIfNotNull(lastPreviewTask, task -> task.cancel(false));
+      long taskDelayTime = CompletionTracker.calcDebounceTime(editor);
```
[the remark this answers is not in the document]
yonip23: nvm, i thought this pr targets debouncing of the rendering, not the completions. you can resolve this

src/main/java/com/tabnine/userSettings/AppSettingsComponent.kt
yairco1990: I want to let us play with this value easily to get the correct default value.
also, I'm not sure this is a bad idea to add it, since some users type faster and some slower, and they can adapt this value to their needs.

src/main/java/com/tabnine/inline/CompletionTracker.kt
yairco1990: https://github.com/codota/tabnine-intellij/pull/376/files#diff-64b8b8695cd691da2eb473d6617e1614f6814291cb0f1623abb32c3154c984eaR13-R14

PR #395, opened 2022-10-13
support debouncing fixed value experiment

yairco1990:


yairco1990:
@yonip23 this is the plugin part of the debounce experiment. on your free time, of course :)

src/main/java/com/tabnine/userSettings/AppSettingsComponent.kt
```diff
@@ -61,13 +62,20 @@ class AppSettingsComponent {
             debounceTimeComponent.isEnabled = false
         }
 
-        panel = FormBuilder.createFormBuilder()
+        val panelBuilder = FormBuilder.createFormBuilder()
             .addLabeledComponent("Log File Path (requires restart): ", logFilePathComponent, 1, false)
             .addLabeledComponent("Log level (requires restart): ", logLevelComponent, 1, false)
-            .addLabeledComponent("Min delay between completions (requires restart): ", debounceTimeComponent, 1, false)
+
+        if (!isFixedDebounceConfigured()) {
```
yonip23, line 69: Mmm I'm a bit worried that there will be some confused users, like if someone already set this value and now it disappears from the ui or something like that.

I guess there's not much we can do, so we just need to remember it. Maybe let Maya know about this change

src/main/java/com/tabnine/inline/DebounceManager.kt
```diff
@@ -0,0 +1,28 @@
...
+    )
+
+    @JvmStatic
+    fun getDebounceInterval(): Long {
+        if (CapabilitiesService.getInstance().isCapabilityEnabled(Capability.DEBOUNCE_VALUE_300)) {
+            return 300
+        }
+        if (CapabilitiesService.getInstance().isCapabilityEnabled(Capability.DEBOUNCE_VALUE_600)) {
+            return 600
+        }
```
yonip23, line 20: i'd extract this logic to a utility file with a static function that returns `Long?`, and then this code can look like this:
```kotlin
return getDebounceMsFromCapabilities() ?? AppSettingsState.instance.debounceTime
```

yonip23 (approved):
lgtm, see comments

src/main/java/com/tabnine/inline/DebounceManager.kt
```diff
@@ -0,0 +1,28 @@
...
+        return AppSettingsState.instance.debounceTime
+    }
+
+    @JvmStatic
+    fun isFixedDebounceConfigured(): Boolean {
```
yonip23, line 25: this function can be extracted to the utility file as well

src/main/java/com/tabnine/userSettings/AppSettingsComponent.kt
yairco1990: you're right, but this is for a limited time (until we determine the default value) and for new users.
we have no more than 20 users right now that use this setting, so I think it's ok..

PR #398, opened 2022-10-23
disable suggestions on paste (length > 1) or tab(indent in)

yairco1990:
Copy of the same behaviour from vscode
https://github.com/codota/tabnine-vscode/pull/956

**before**:
https://user-images.githubusercontent.com/10290661/197491449-2db715aa-5958-42ba-becf-756bbcb20ea0.mp4
https://user-images.githubusercontent.com/10290661/197491456-fbf35226-231a-49f1-aad9-3ea1d69261e9.mp4

**after**:
https://user-images.githubusercontent.com/10290661/197491512-1b9ded67-9a19-4e33-80c7-38fa8f49fa59.mp4
https://user-images.githubusercontent.com/10290661/197491502-24e8c81a-c73e-4663-9856-8044b1b5ab7b.mp4

### DIFF
--- a/src/main/java/com/tabnine/inline/TabnineDocumentListener.java
+++ b/src/main/java/com/tabnine/inline/TabnineDocumentListener.java
@@ -8,15 +8,12 @@ import com.intellij.ide.DataManager;
 import com.intellij.openapi.actionSystem.CommonDataKeys;
 import com.intellij.openapi.actionSystem.DataContext;
 import com.intellij.openapi.application.ApplicationManager;
-import com.intellij.openapi.diagnostic.Logger;
 import com.intellij.openapi.editor.Document;
 import com.intellij.openapi.editor.Editor;
 import com.intellij.openapi.editor.EditorKind;
 import com.intellij.openapi.editor.event.BulkAwareDocumentListener;
 import com.intellij.openapi.editor.event.DocumentEvent;
-import com.intellij.openapi.util.TextRange;
 import com.intellij.openapi.wm.IdeFocusManager;
-import com.intellij.util.DocumentUtil;
 import com.tabnine.capabilities.SuggestionsModeService;
 import com.tabnine.general.EditorUtils;
 import java.awt.*;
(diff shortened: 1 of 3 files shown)


src/main/java/com/tabnine/inline/TabnineDocumentListener.java
```diff
@@ -73,27 +69,7 @@ private boolean shouldIgnoreChange(DocumentEvent event, Editor editor, int offse
       return true;
     }
 
-    if (!CompletionUtils.isValidMidlinePosition(document, offset)) {
-      return true;
-    }
-
-    return isInTheMiddleOfWord(document, offset);
```
yairco1990, line 80: this check is no longer necessary since the isValidMidlinePosition already checks this case

src/main/java/com/tabnine/inline/TabnineDocumentListener.java
```diff
@@ -42,10 +39,9 @@ public void documentChangedNonBulk(@NotNull DocumentEvent event) {
         .invokeLater(
             () -> {
               int offset =
-                  editor.getCaretModel().getOffset()
-                      + (ApplicationManager.getApplication().isUnitTestMode()
-                          ? event.getNewLength()
-                          : 0);
+                  ApplicationManager.getApplication().isUnitTestMode()
+                      ? event.getOffset() + event.getNewLength()
+                      : editor.getCaretModel().getOffset();
```
yonip23, line 44: is this a bugfix? how did this affect us in tests and in production?

yonip23 (approved):
if you have a demo video it'd be nice to share it here on the pr

src/main/java/com/tabnine/inline/TabnineDocumentListener.java
yairco1990: I just changed the value in the test mode. the value in production stays the same.
it seems to be an issue with the previous value in the tests. 
Whoever wrote this line of code relies on the fact that `editor.getCaretModel().getOffset();` returns the offset after the doc change in the running mode, and in the test mode it returns the offset before the doc change.
I experienced some inconsistency with this value in the test mode in some cases (sometimes returns the old offset and sometimes the new offset), so I changed it to this one which should be consistent.

PR #409, opened 2022-11-15
DEV2-1681 - add "Getting Started" page

yairco1990:
**Status bar changes:**
The previous status bar widget was a pure swing component that was very hard to manage, and the popup wasn't supported.
I changed it to extend `StatusBarWidget.MultipleTextValuesPresentation` instead of the `CustomStatusBar` so we can support the popup easily, add a separate method that gets the icon, and a separate method that gets the text.

**Open the getting-started page:**
I used CEF, which is chromium for java. the issue is that only IJ version 20.2 and later support this feature, so I used reflection to get over the compilation error for the older versions.

**QA:**

Checked that the popup and the buttons work properly:
https://user-images.githubusercontent.com/10290661/202397396-db818774-68b6-4c2f-9b3d-281d88f1dc56.mp4

Checked that when there are **no** local binary versions, the getting-started page should open on IDE startup:
https://user-images.githubusercontent.com/10290661/202398011-a533f138-d3b7-4157-ae46-4029ec034232.mp4

Checked that when there are local binary versions, the getting-started page should not open on IDE startup:
https://user-images.githubusercontent.com/10290661/202398252-1bef4d53-84d4-45d5-b2e8-f40370bb9dac.mp4


src/main/java/com/tabnine/general/GettingStartedManager.kt
```diff
@@ -0,0 +1,102 @@
...
+        return try {
+            getOpenEditorMethod()
+            true
+        } catch (e: Exception) {
+            false
+        }
+    }
+
+    @JvmStatic
+    fun getOpenEditorMethod(): Method {
```
yairco1990, line 83: workaround for unsupported IDEs

src/main/java/com/tabnine/binary/fetch/BootstrapperSupport.java
```diff
@@ -14,9 +16,13 @@ static Optional<BinaryVersion> bootstrapVersion(
       BundleDownloader bundleDownloader) {
     Optional<BinaryVersion> localBootstrapVersion =
         locateLocalBootstrapSupportedVersion(localBinaryVersions);
-    return localBootstrapVersion.isPresent()
-        ? localBootstrapVersion
-        : downloadRemoteVersion(binaryRemoteSource, bundleDownloader);
+    if (localBootstrapVersion.isPresent()) {
+      return localBootstrapVersion;
+    }
+    if (!ApplicationManager.getApplication().isUnitTestMode()) {
```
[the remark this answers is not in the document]
yairco1990: `PropertiesComponent.getInstance()` returns `null` in unit test mode. I'll try to find a way to mock it

src/main/java/com/tabnine/general/GettingStartedManager.kt
```diff
@@ -0,0 +1,102 @@
...
+import java.util.Arrays
+import java.util.concurrent.TimeUnit
+import javax.swing.SwingConstants
+
+const val PAGE_TITLE = "Tabnine - Getting Started"
+const val PAGE_URL = "https://www.tabnine.com/getting-started/intellij?origin=ide"
+const val IS_GETTING_STARTED_OPENED_KEY = "is-getting-started-opened"
+const val AFTER_PROJECT_OPENED_DELAY_SECONDS = 15L
+
+object GettingStartedManager {
```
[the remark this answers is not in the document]
yairco1990: I didn't know functions could be at the top level 😳

src/main/java/com/tabnine/binary/fetch/BootstrapperSupport.java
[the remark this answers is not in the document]
yonip23: no need, that's fine
although i'd like to see dedicated tests to this new component (as I mentioned above)

src/main/java/com/tabnine/general/GettingStartedManager.kt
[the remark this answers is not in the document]
yonip23: lol yeah, its practically a static function

src/main/java/com/tabnine/general/GettingStartedManager.kt
```diff
@@ -0,0 +1,102 @@
...
+                fileEditorManagerEx.currentWindow?.closeFile(currentFocusedFile)
+            }
+        } catch (e: Exception) {
+            BrowserUtil.browse(PAGE_URL)
+        }
+        markPageAsShown()
+    }
+
+    @JvmStatic
+    fun isInIdeWebPageSupported(): Boolean {
```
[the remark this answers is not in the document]
yairco1990: the `openPageOnProject` method needs to be independent, so I don't think it should get it from outside

src/main/java/com/tabnine/general/GettingStartedManager.kt
```diff
@@ -0,0 +1,92 @@
...
+    if (!isPageShown()) {
+        executeUIThreadWithDelay(
+            { openPageOnAllProjects() },
+            AFTER_PROJECT_OPENED_DELAY_SECONDS,
+            TimeUnit.SECONDS
+        )
+    }
+}
+
+fun openPageOnAllProjects() {
```
yonip23, line 30: mark all non public methods with `private`

src/main/java/com/tabnine/general/GettingStartedManager.kt
```diff
@@ -0,0 +1,92 @@
...
+    if (isInIdeWebPageSupported()) {
```
yonip23, line 31: no need for this `if` now that you `try/catch` inside the function

src/main/java/com/tabnine/general/GettingStartedManager.kt
```diff
@@ -0,0 +1,92 @@
...
+            project,
+            PAGE_TITLE,
+            PAGE_URL,
+            null
+        )
+        // close the duplicated file in the new editor and set the web page to be the single tab
+        if (currentFocusedFile != null) {
+            fileEditorManagerEx.currentWindow?.closeFile(currentFocusedFile)
+        }
+    } catch (e: Exception) {
```
yonip23, line 61: i wanted to say "add a comment explaining that this catch means that the IDE doesn't support this feature", but even better would be to create a custom exception `NotSupportedByIDEVersion`:
```kotlin
class NotSupportedByIDEVersion: Excpetion("Operaion is not supported by this IDE version")
```
(or something along the lines)

src/main/java/com/tabnine/general/GettingStartedManager.kt
```diff
@@ -0,0 +1,92 @@
...
+        BrowserUtil.browse(PAGE_URL)
+    }
+    markPageAsShown()
+}
+
+fun isInIdeWebPageSupported(): Boolean {
```
yonip23, line 67: remove this method

yonip23 (commented):
please explain here shortly the changes in the status bar area

src/main/java/com/tabnine/general/GettingStartedManager.kt
yairco1990: ok. I just thought it will be clearer that way..

yonip23 (approved):
🚀

src/main/java/com/tabnine/binary/fetch/BootstrapperSupport.java
```diff
@@ -14,9 +16,14 @@ static Optional<BinaryVersion> bootstrapVersion(
...
         locateLocalBootstrapSupportedVersion(localBinaryVersions);
-    return localBootstrapVersion.isPresent()
-        ? localBootstrapVersion
-        : downloadRemoteVersion(binaryRemoteSource, bundleDownloader);
+    if (localBootstrapVersion.isPresent()) {
+      return localBootstrapVersion;
+    }
+    if (ApplicationManager.getApplication() != null
+        && !ApplicationManager.getApplication().isUnitTestMode()) {
+      GettingStartedManagerKt.handleFirstTimePreview();
```
dimacodota, line 24: no for now, but I suggest refactoring this, this dependency is really weird here, not sure I understand how version fetching is related to getting started...  maybe raise here some "just-installed" event here and subscribe to it somewhere

dimacodota (approved):
lgtm

PR #417, opened 2022-12-03
Dev2 1835 - do not modify the old suffix in case multiline suggestions

dimacodota:
**problem:**  
Tabnine returns the  `old_suffix`  to be replaced by the new_prefix (current suggestion)
for now, this logic is relevant only in single-line suggestions.

**example:** 
 in the case of `console("H<caret>")`  and `new_prefix :ello")` Tabnine will return  `old_suffix: ")` to be replaced by the `new_prefix`

**notes:**
 - this logic s not relevant to multiline suggestions because it is already handled [here](https://github.com/codota/triton-cloud-frontend/blob/a8997be6f66c7312567b819068e0f05db798afdb/app/suggestion_processing/postprocessors/overlap.py#L18)
 - we have the same workaround in vscode as well https://github.com/codota/tabnine-vscode/blob/c0a9a7d5a19e4a9ef23579ce8688d216d1aab439/src/provideInlineCompletionItems.ts#L87
 - this logic is calculated [here](https://github.com/codota/acpl-paid/blob/13bfa7ffde61209300ac7167505c211727b93995/tabnine/server/src/postprocess.rs#L166)
and because it is used by the Ycmd mode, for now, I've decided to handle it on the plugin level, moving forward  we should consider removing it from the binary

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -89,3 +89,5 @@ jobs:
         uses: snyk/actions/gradle-jdk11@master
         env:
           SNYK_TOKEN: ${{ secrets.SNYK_TOKEN }}
+        with:
+          args: --severity-threshold=high
--- a/src/main/java/com/tabnine/inline/CompletionPreviewUtils.kt
+++ b/src/main/java/com/tabnine/inline/CompletionPreviewUtils.kt
@@ -9,6 +9,6 @@ fun hadSuffix(currentCompletion: TabNineCompletion): Boolean {
 fun isSingleLine(currentCompletion: TabNineCompletion): Boolean {
     return !currentCompletion.getSuffix().trim()?.contains("\n") ?: true
 }
-fun shouldRemoveSuffixIfSingleLine(currentCompletion: TabNineCompletion): Boolean {
+fun shouldRemoveSuffix(currentCompletion: TabNineCompletion): Boolean {
     return hadSuffix(currentCompletion) && isSingleLine(currentCompletion)
 }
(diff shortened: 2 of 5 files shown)


src/main/java/com/tabnine/inline/CompletionPreviewUtils.kt
```diff
@@ -5,3 +5,10 @@ import com.tabnine.prediction.TabNineCompletion
 fun hadSuffix(currentCompletion: TabNineCompletion): Boolean {
     return currentCompletion.oldSuffix?.trim()?.isNotEmpty() ?: false
 }
+
+fun isSingleLine(currentCompletion: TabNineCompletion): Boolean {
+    return !currentCompletion.getSuffix().trim()?.contains("\n") ?: true
+}
+fun shouldRemoveSuffixIfSingleLine(currentCompletion: TabNineCompletion): Boolean {
```
yonip23, line 12: Remove "IfIsSingleLine" from the name - it's an internal detail to this function and irrelevant to the caller

yonip23 (approved):
Lgtm
Not sure why you suggest removing the logic completely from the binary though

dimacodota:
> Lgtm Not sure why you suggest removing the logic completely from the binary though
my suggestion was to remove the 'old_suffix` initialization in case of multiline, the caller should not care whether the suggestion is multiline or not

.github/workflows/ci.yml
```diff
@@ -89,3 +89,5 @@ jobs:
         uses: snyk/actions/gradle-jdk11@master
         env:
           SNYK_TOKEN: ${{ secrets.SNYK_TOKEN }}
+        with:
+          args: --severity-threshold=high
```
dimacodota, line 93: @yonip23  - FYI

PR #423, opened 2022-12-07
DEV2-1383 add auto import toggle

yonip23:


yairco1990 (approved):
lgtm
are we sure we want to change the default behavior?

yonip23:
That's what I'm trying to do lol
Will make sure it's true by default of course

yonip23:
@yairco1990 @dimacodota I removed the xml settings file from my machine, restarted the IDE and made sure that the settings panel opens with this checkbox selected, and that the auto import actually works.
This is the new text:
image

dimacodota (approved):
lgtm

PR #425, opened 2022-12-11
DEV2-1865 add binaries location settings option

yonip23:
image

⚠️ Note - there's no input validation atm, so the path change can lead to unexpected behavior

### DIFF
--- a/src/main/java/com/tabnine/general/StaticConfig.java
+++ b/src/main/java/com/tabnine/general/StaticConfig.java
@@ -2,9 +2,12 @@ package com.tabnine.general;
 
 import static java.awt.Color.decode;
 
+import com.intellij.openapi.application.Application;
+import com.intellij.openapi.application.ApplicationManager;
 import com.intellij.openapi.extensions.PluginId;
 import com.intellij.openapi.util.IconLoader;
 import com.intellij.openapi.util.SystemInfo;
+import com.intellij.util.ObjectUtils;
 import com.intellij.util.text.SemVer;
 import com.tabnine.binary.exceptions.InvalidVersionPathException;
 import com.tabnine.userSettings.AppSettingsState;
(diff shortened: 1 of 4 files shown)


src/main/java/com/tabnine/general/StaticConfig.java
```diff
@@ -165,6 +168,24 @@ private static String getDistributionName() {
...
   public static Path getBaseDirectory() {
+    // Unit tests don't initialize the application, so `ApplicationManager.getApplication` will
+    // return null.
+    Boolean isUnitTest =
+        ObjectUtils.doIfNotNull(ApplicationManager.getApplication(), Application::isUnitTestMode);
+    if (isUnitTest == null || isUnitTest) {
+      return getDefaultBaseDirectory();
+    }
+
+    String binariesFolderOverride = AppSettingsState.getInstance().getBinariesFolderOverride();
```
yairco1990, line 179: does `AppSettingsState` is null in unit tests?

yairco1990 (approved):
lgtm

src/main/java/com/tabnine/general/StaticConfig.java
yonip23: Yes, because it is a registered service, and since the application is not initialized so are it's services

yairco1990: so why do the unit tests test this function? 🤔

yonip23: They aren't. 
They use this function for the binary downloader tests and such, but before this pr, this function wasn't dependent on the application so it all worked fine.
Now that this function reads from the settings, it became dependent- so it needs a special handling

dimacodota (approved):
lgtm

PR #424, opened 2022-12-08
Dev2 1536 add logs and increase threshold

yonip23:
I increased the TO from 20 to 80, because this 20 threshold was set a long time ago - long before we started polling the binary every 10 seconds.
I decided on 80 because with the logs I added, without doing anything in the IDE it took ~3min to reach 58 requests:
```
2022-12-08 16:44:50,235 [   6742]  DEBUG - BinaryProcessRequesterProvider - <<ALPHA LOG>> Called successfulRequest with request #1 
...
2022-12-08 16:47:30,269 [ 166776]  DEBUG - BinaryProcessRequesterProvider - <<ALPHA LOG>> Called successfulRequest with request #58 
```
and when I started typing fast immediately after restarting the IDE, I got to 80 in < 30 seconds:
```
2022-12-08 17:09:02,870 [   6688]  DEBUG - BinaryProcessRequesterProvider - <<ALPHA LOG>> Called successfulRequest with request #1 
...
2022-12-08 17:09:31,874 [  35692]  DEBUG - BinaryProcessRequesterProvider - <<ALPHA LOG>> Called successfulRequest with request #80 
```

So I'm guessing the 80 threshold is ok for avg. users - lmk wdyt

PR #426, opened 2022-12-11
DEV2-1716 previous and next suggestion events

yonip23:
This PR also includes a fix to the tests from #424 - since this PR increased the TOs threshold to 80, it caused the tests to take longer and even fail sometimes.
I fixed it by passing the TOs and restarts thresoleds as data members to `BinaryProcessRequesterProvider`, and change the values for tests in `DependencyContainer.setTesting`

### DIFF
--- a/src/main/java/com/tabnine/inline/CompletionPreview.java
+++ b/src/main/java/com/tabnine/inline/CompletionPreview.java
@@ -34,6 +34,8 @@ public class CompletionPreview implements Disposable {
 
   private final CompletionPreviewListener previewListener =
       DependencyContainer.instanceOfCompletionPreviewListener();
+  private final CompletionPreviewToggleEventSender completionPreviewToggleEventSender =
+      DependencyContainer.instanceOfCompletionPreviewToggleEventSender();
 
   public final Editor editor;
   private TabnineInlay tabnineInlay;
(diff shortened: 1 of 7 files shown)


src/main/java/com/tabnine/inline/CompletionPreviewToggleEventSender.kt
```diff
@@ -0,0 +1,19 @@
+package com.tabnine.inline
+
+import com.intellij.openapi.application.ApplicationManager
+import com.tabnine.binary.BinaryRequestFacade
+import com.tabnine.binary.requests.analytics.EventRequest
+
+class CompletionPreviewToggleEventSender(private val binaryRequestFacade: BinaryRequestFacade) {
```
yonip23, line 7: this is the main change that's not related to the tests

src/main/java/com/tabnine/inline/CompletionPreview.java
```diff
@@ -91,6 +93,7 @@ public void togglePreview(CompletionOrder order) {
     tabnineInlay = TabnineInlay.create(this);
 
     createPreview();
+    completionPreviewToggleEventSender.sendToggleEvent(order, currentIndex);
```
yonip23, line 96: this is the main change that's not related to the tests

dimacodota (approved):
lgtm

PR #427, opened 2022-12-11
DEV2-1717 manual suggestion trigger event

yonip23:
I also added tests to this event + prev/next events,
and also added an "escape" event with a test


src/main/java/com/tabnine/intellij/completions/TabNineCompletionContributor.java
```diff
@@ -94,6 +101,22 @@ public void fillCompletionVariants(
     resultSet.addAllElements(createCompletions(completions, parameters, resultSet));
   }
 
+  /**
+   * An ugly hack I (Yoni) found during debug - it appears to be the case that `invocationCount`
+   * flips to 1 when you trigger a popup completion with ctrl+space, remains 1 as long as this popup
+   * remains open, and is set to 0 if it was auto triggered. So the hack is to keep track on when it
+   * changes from 0 to non-zero - because this means that this is the first manual invocation of the
+   * current popup menu.
+   */
```
yonip23, line 110: @dimacodota @amircodota wdyt?
In vscode we have a `CompletionContext` which has this data (whether or not the completion was manually triggered), but here I couldn't find such information - I looked up their docs and debugged the area for a bit.
If you can find any relevant piece of information that'd be delightful, but if not - wdyt about this hack? is it in the "ok"-ish realm?

amircodota: I'm fine with this hack, if there's no other way.

But are we sure there's no other way?

I see parameters have an `isAutoPopup` method. Does that not work?

amircodota: :partying_face:

yonip23: @amircodota @dimacodota im using it now, and I see that if I `ctrl+space` to get the popup, and then while its open I keep typing - it counts the next keystrokes as a manual popup each. This is until the popup closes.
I think my implementation is a bit more accurate for our use case, but I still prefer to use JB's `isAutoPopup` method. Wdyt?

amircodota (approved):
LGTM

Would be nice to have a test for this

dimacodota (approved):
lgtm

PR #429, opened 2022-12-12
DEV2-1536 binary TOs - measure time instead of counter

yonip23:
### Problem
We were declaring the binary as "bad" based on x consecutive requests that timed out.
This approach proved to be incorrect, as since we defined this x, we added some serious polling requests - thus we began to reach this timeout threshold too quickly.

### Solution
In #424, I increased this threshold from 20 to 80, just to remediate with a quick and easy solution.

This pr is the more serious solution - measure time between successful requests instead of TOs counter.
The logic goes like so:
1. on successful request - restart timeouts tracking
2. on timeout:
    1. if it's the first after a successful request - set the time to now.
    2. if now - `firstTimeoutMillis` > threshold -> restart
Current threshold is 1 minute (60k ms).

### DIFF
--- a/src/main/java/com/tabnine/binary/BinaryProcessRequesterProvider.java
+++ b/src/main/java/com/tabnine/binary/BinaryProcessRequesterProvider.java
@@ -18,24 +18,24 @@ public class BinaryProcessRequesterProvider {
   private final BinaryProcessGatewayProvider binaryProcessGatewayProvider;
 
   private int consecutiveRestarts = 0;
-  private int consecutiveTimeouts = 0;
+  private Long firstTimeoutTimestamp = null;
   private BinaryProcessRequesterPoller poller;
   private BinaryProcessRequester binaryProcessRequester;
   private Future<?> binaryInit;
   private AtomicInteger requestsCounter = new AtomicInteger(0);
-  private final int timeoutsThreshold;
+  private final int timeoutsThresholdMillis;
   private final int restartsThreshold;
 
   private BinaryProcessRequesterProvider(
       BinaryRun binaryRun,
       BinaryProcessGatewayProvider binaryProcessGatewayProvider,
       BinaryProcessRequesterPoller poller,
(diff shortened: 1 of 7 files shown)


yairco1990 (approved):
pagaz

src/main/java/com/tabnine/binary/BinaryProcessRequesterProvider.java
```diff
@@ -105,7 +105,12 @@ public void onTimeout() {
 
     Logger.getInstance(getClass()).info("TabNine's response timed out.");
 
-    if (++consecutiveTimeouts >= timeoutsThreshold) {
+    long now = System.currentTimeMillis();
+    if (firstTimeoutTimestamp == null) {
+      firstTimeoutTimestamp = now;
```
assapir, line 110: You don't want to set it back to null on a good request?

src/main/java/com/tabnine/binary/BinaryProcessRequesterProvider.java
```diff
@@ -82,7 +82,7 @@ public void onDead(Throwable e) {
     Logger.getInstance(BinaryProcessRequesterProvider.class)
         .debug(String.format("<<ALPHA LOG>> %s", msg));
 
-    consecutiveTimeouts = 0;
+    firstTimeoutTimestamp = null;
     Logger.getInstance(getClass()).warn("Tabnine is in invalid state, it is being restarted.", e);
```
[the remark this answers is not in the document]
yonip23: No, I restart timeouts tracking when the binary is declared dead and being restarted

src/main/java/com/tabnine/binary/BinaryProcessRequesterProvider.java
yonip23: I do it

dimacodota (approved):

PR #432, opened 2022-12-13
DEV2-1779: fix caret listener race condition and change SuggestionShownRequest response handling

yairco1990:
- fix caret listener race condition
- change SuggestionShownRequest response handling so it will ignore the response type

https://user-images.githubusercontent.com/10290661/207381438-486bf243-fd26-4389-96e4-41d53ed1f2a1.mp4

### DIFF
--- a/src/main/java/com/tabnine/inline/listeners/InlineCaretListener.kt
+++ b/src/main/java/com/tabnine/inline/listeners/InlineCaretListener.kt
@@ -27,7 +27,7 @@ class InlineCaretListener(private val completionPreview: CompletionPreview) : Ca
     }
 
     private fun isSingleOffsetChange(event: CaretEvent): Boolean {
-        return event.oldPosition.column + 1 == event.newPosition.column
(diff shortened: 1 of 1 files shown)


src/main/java/com/tabnine/inline/listeners/InlineCaretListener.kt
```diff
@@ -11,18 +11,25 @@ import com.tabnine.inline.InlineCompletionCache
...
+        if (isSingleOffsetChange(event)) {
+            return
+        }
 
         Disposer.dispose(completionPreview)
         InlineCompletionCache.instance.clear(event.editor)
     }
 
+    private fun isSingleOffsetChange(event: CaretEvent): Boolean {
+        return event.oldPosition.column + 1 == event.newPosition.column
```
yonip23, line 30: add row check as well, because you can move from (x, y) to (x+1,y+1) and you'll mistakenly return true in this case

yonip23 (changes requested):

yonip23 (approved):

dimacodota (approved):
LGTM

PR #436, opened 2022-12-14
DEV2-1921 add bump version action

yonip23:
This is a copy of `production.yml`, but with the added functionality of setting the version from an input to the action.
I'm not changing the original action on purpose - since I can't test this job before it is merged, I don't want to screw the existing action 🙂


yonip23:
@dimacodota im afraid it worn't work until you add a `CODOTA_BUILDER_TOKEN` secret to the repos https://github.com/codota/acpl-paid/blob/fbddc07f9174b3b87ea45d3efc5b68153c9d6d46/.github/workflows/BinariesReleaseVersion.yaml#L19

.github/workflows/bump-version.yml
```diff
@@ -0,0 +1,48 @@
+name: Bump Version
+
+on:
+  workflow_dispatch:
+    inputs:
+      version:
```
dimacodota, line 6: why not use the existing version++ by default, and leave it as optional

.github/workflows/bump-version.yml
```diff
@@ -0,0 +1,48 @@
...
+          set -ex
+          git config --global user.email "builder+skip-ci@codota.com"
+          git config --global user.name "codota-builder"
+
+      - name: Commit changes
+        run: |
+          NEW_VERSION="${{ github.event.inputs.version }}"
+          git pull origin master --no-edit
+          git commit -am "Set meta version to $NEW_VERSION"
+          git push
```
dimacodota, line 48: you need the tag handling

dimacodota: i mean you probably want to create and push the tag

yonip23: no, im leaving this to the production release action - it creates the tag and all.
i just want this to work as a first step, then we can later merge both actions -wdyt?

yonip23: the issue is that if i break the existing production release on master its gonna be a pain in my ass to fix, since I don't have permissions.
so each change will require a pr with merge approve and all that.
that's why I'm creating a separate action that doesn't change the existing one

.github/workflows/production-new-version.yml
```diff
@@ -0,0 +1,109 @@
...
+      - name: Commit changes
+        run: |
+          NEW_VERSION="${{ steps.get_build_version.outputs.build_version }}"
+          git pull origin master --no-edit
+          git commit -am "Set version to $NEW_VERSION"
+          git push
+          git tag v$NEW_VERSION
+          git push origin v$NEW_VERSION
+
+      - name: Create Tag
```
dimacodota, line 75: why is it dupliacate?

yonip23: because i forgot to remove it of course 😬

yonip23:
@dimacodota what's up with this? can you approve?

dimacodota (approved):
lgtm

PR #446, opened 2022-12-25
DEV2-1802: handle connection lost

yairco1990:
binary [PR](https://github.com/codota/acpl-paid/pull/2440)

https://user-images.githubusercontent.com/10290661/209455149-0b8b5cb3-a3b9-453f-8245-f2cbf1b0f0b2.mp4


src/main/java/com/tabnine/general/Utils.java
```diff
@@ -92,6 +92,14 @@ public static long getDaysDiff(Date date1, Date date2) {
     return -1;
   }
 
+  public static long getHoursDiff(Date date1, Date date2) {
+    if (date1 != null && date2 != null) {
+      return TimeUnit.HOURS.convert(
+          Math.abs(date2.getTime() - date1.getTime()), TimeUnit.MILLISECONDS);
```
yonip23, line 98: i think its better to account for order, i.e don't calculate with `abs`

src/main/java/com/tabnine/general/StaticConfig.java
```diff
@@ -114,14 +123,38 @@ public static String getTabNineBetaVersionUrl() {
         .orElse(getServerUrl() + "/beta_version");
   }
 
-  public static Icon getTabnineIcon(@Nullable ServiceLevel serviceLevel) {
+  public static SubscriptionType getSubscriptionType(@Nullable ServiceLevel serviceLevel) {
```
yonip23, line 126: I think that this function should be extracted to a separate class. and add a function `getTabnineIcon` to the `SubscriptionType` enum, such that each enum variant will return its own icon (instead of implementing this with an external switch-case

src/main/java/com/tabnine/notifications/ConnectionLostNotificationHandler.kt
```diff
@@ -0,0 +1,61 @@
...
+import java.util.concurrent.atomic.AtomicBoolean
+
+private const val INTERVAL_BETWEEN_NOTIFICATIONS_HOURS = 5
+private const val NOTIFICATION_CONTENT =
+    "<b>Tabnine lost internet connection.</b>" +
+        "<br/>" +
+        "If your internet is connected and you're seeing this message, contact Tabnine support."
+
+class ConnectionLostNotificationHandler {
+    private val ourGroup = NotificationGroup(StaticConfig.BRAND_NAME, NotificationDisplayType.STICKY_BALLOON, false)
```
yonip23, line 26: maybe extract to const in `StaticConfig`? it's being used in 2 places now if I recall

src/main/java/com/tabnine/notifications/ConnectionLostNotificationHandler.kt
```diff
@@ -0,0 +1,61 @@
...
+    private var lastNotificationTime: Date? = null
+    private var isRegistered = AtomicBoolean(false)
+
+    fun handleConnectionLostEvent() {
+        if (!isRegistered.getAndSet(true)) {
```
yonip23, line 31: you know me... invert condition 🙂 
im not commenting this on every `if`, but here you've got like 4 more indents down the road, please spare as one lol

src/main/java/com/tabnine/statusBar/StatusBarEmptySymbolGenerator.java
```diff
@@ -0,0 +1,20 @@
...
+
+import java.util.Arrays;
+import java.util.List;
+
+// Workaround to address the issue where IJ does not render the width of the status bar icon
+// correctly when the icon changes
+// because it doesn't detect a change in the widget text (it renders it correctly just after
+// document change occurs).
+// Creating a different empty text each time will solve the problem.
+public class StatusBarEmptySymbolGenerator {
```
yonip23, line 11: this is very bote 🤮 
if there's really nothing else that solves it... 🤷🏻‍♂️ 
I'd try to contact jb's support about this anyway

yonip23, line 11: also, please re-implement in kotlin, and this can be a global function without a class.

src/test/java/com/tabnine/integration/ConnectionLostNotificationTests.java
```diff
@@ -0,0 +1,86 @@
...
+            Notifications.TOPIC,
+            new Notifications() {
+              @Override
+              public void notify(@NotNull Notification notification) {
+                assertTrue(notification.getContent().contains("Tabnine lost internet connection"));
+              }
+            });
+
+    new ConnectionLostNotificationHandler().handleConnectionLostEvent();
+    notifyStateChangedTopic(getMockedStateResponse(CloudConnectionHealthStatus.Failed));
```
yonip23, line 33: i guess you should wait here for the notification to propagate to the listener, no?

src/test/java/com/tabnine/integration/ConnectionLostNotificationTests.java
```diff
@@ -0,0 +1,86 @@
...
+  public void should_show_connection_lost_notification_when_connection_unhealthy() {
+    ApplicationManager.getApplication()
+        .getMessageBus()
+        .connect()
+        .subscribe(
+            Notifications.TOPIC,
+            new Notifications() {
+              @Override
+              public void notify(@NotNull Notification notification) {
+                assertTrue(notification.getContent().contains("Tabnine lost internet connection"));
```
yonip23, line 28: extract const, or even better - use the actual message const if possible

src/test/java/com/tabnine/integration/ConnectionLostNotificationTests.java
```diff
@@ -0,0 +1,86 @@
...
+              public void notify(@NotNull Notification notification) {
+                if (notification.getContent().contains("Tabnine lost internet connection")) {
+                  numOfNotificationAppearance.incrementAndGet();
+                }
+              }
+            });
+
+    new ConnectionLostNotificationHandler().handleConnectionLostEvent();
+    for (int i = 0; i < 10; i++) {
+      notifyStateChangedTopic(getMockedStateResponse(CloudConnectionHealthStatus.Failed));
```
yonip23, line 55: extract to function `notifyHealthStatus(status, times)` and use in all tests

src/test/java/com/tabnine/integration/StatusBarWidgetTests.java
```diff
@@ -0,0 +1,29 @@
+package com.tabnine.integration;
```
yonip23, line 1: this is only health-related status bar tests, so either rename the file or test the entire status bar behavior 😈

yonip23 (commented):
great work and well done for testing!
had some comments i'd like you to refer to before merging - but not blocking
are you sure we need all these icons? just double check if you didn't upload any additional ones lol

src/main/java/com/tabnine/statusBar/StatusBarEmptySymbolGenerator.java
yairco1990: great. believe me, I tried anything. nothing solves this.

src/test/java/com/tabnine/integration/ConnectionLostNotificationTests.java
yairco1990: I'm not sure if this is a good practice to mark a variable as public only for test purposes. wdyt?

src/main/java/com/tabnine/notifications/ConnectionLostNotificationHandler.kt
yairco1990: totally forgot you hate it 😂

src/main/java/com/tabnine/statusBar/StatusBarEmptySymbolGenerator.java
yairco1990: why not a class?

src/test/java/com/tabnine/integration/ConnectionLostNotificationTests.java
yairco1990: no, it's a synchronous code. the listener will be executed before the assert.

src/main/java/com/tabnine/general/StaticConfig.java
```diff
@@ -74,8 +74,12 @@ public class StaticConfig {
   public static final long BINARY_PROMOTION_POLLING_DELAY = 10_000L; // 10 seconds
 
   public static final String OPEN_HUB_ACTION = "OpenHub";
-  private static final Set<ServiceLevel> PRO_SERVICE_LEVELS =
-      EnumSet.of(ServiceLevel.PRO, ServiceLevel.TRIAL);
+
+  // As far as I understand, This registers `BRAND_NAME` display id as a "sticky balloon" type,
+  // so that when we create a notification with this display id it knows to make it a sticky
+  // balloon.
+  public static NotificationGroup ourGroup =
```
yonip23, line 81:
```suggestion
  public static NotificationGroup OUR_NOTIFICATION_GROUP =

File: src/main/java/com/tabnine/statusBar/StatusBarEmptySymbolGenerator.kt [hunk 11]
```diff
@@ -1,20 +1,17 @@
...
 // correctly when the icon changes
 // because it doesn't detect a change in the widget text (it renders it correctly just after
 // document change occurs).
 // Creating a different empty text each time will solve the problem.
-public class StatusBarEmptySymbolGenerator {
-  private static final String EMPTY_SYMBOL = "\u0000";
-  private static final List<String> EMPTY_SYMBOLS =
-      Arrays.asList(EMPTY_SYMBOL, EMPTY_SYMBOL + EMPTY_SYMBOL);
-  private int currentSymbolIndex = 0;
+class StatusBarEmptySymbolGenerator {
```
yonip23, line 8: you don't need a class with a companion object here - you can declare top level static consts and functions in Kotlin - so you just expose a utility function of `getEmptySimbol`

src/test/java/com/tabnine/testUtils/HealthCheckTestUtils.java
```diff
@@ -0,0 +1,23 @@
+package com.tabnine.testUtils;
+
+import com.intellij.openapi.application.ApplicationManager;
+import com.tabnine.binary.requests.config.CloudConnectionHealthStatus;
+import com.tabnine.binary.requests.config.StateResponse;
+import com.tabnine.lifecycle.BinaryStateChangeNotifier;
+
+public class HealthCheckTestUtils {
+
+  public static void notifyHealthStatusLoop(
```
yonip23, line 10: 1. you can have this function to be exactly as the other one - `notifyHealthStatus` (function overloading)
2. why Java? we prefer new files to be written in Kotlin

yonip23 (approved):
beautiful! had some small comments, but lgtm nonetheless

src/main/java/com/tabnine/statusBar/StatusBarEmptySymbolGenerator.java
yonip23: why have a class when you dont have to 🙄 
that's my personal preference - you don't have to do it if you think otherwise, I approved you pr anyway 😗

src/main/java/com/tabnine/statusBar/StatusBarEmptySymbolGenerator.kt
yairco1990: the problematic use case here is that two components use this static object, and then the first one changes it, the second one changes it back, and then the first one calls it and gets the same symbol as before which will lead to the component not being rendered.

src/main/java/com/tabnine/general/StaticConfig.java
yairco1990: ohh right :)

src/main/java/com/tabnine/statusBar/StatusBarEmptySymbolGenerator.kt
yonip23: oh so each one maintains an instance - I see.
sababa

dimacodota (approved):
lgtm

PR #447, opened 2022-12-27
DEV2-1779: fix late registration to caret listener when debounce is enabled

yairco1990:
when debounce is enabled, the completion preview creation is delayed, and therefore the registration to the caret listener is also delayed as it's attached to the preview.
I added a check that allows rendering completions just in case the offset on the document change and the offset on the rendering moment are equal, otherwise, cancel the rendering.

before fix:

https://user-images.githubusercontent.com/10290661/209736592-464ee789-6eaa-431c-a8b1-8215fef01d1e.mp4

after fix:

https://user-images.githubusercontent.com/10290661/209736609-026d7345-7b9f-4dc3-b8e3-fd17527b23e5.mp4


dimacodota (approved):
LGTM

PR #449, opened 2022-12-29
DEV2-2041: fix impact of multiple document change event on inline completions

yairco1990:
**Background:**
In some cases, IJ fires multiple events on a single document change. For example, after a line breakת it fires two events, where the first event is relevant for the completion but the other one is not ("\n" change event and then "empty" event with offset correction).
When we are about to render the completion, we check if it's relevant by checking that modificationStamp and offset haven't changed. 

**The problem:**
When such a scenario occurs (like that in the example), This is the order of events:
1. we get the first "\n" event 
2. we save the editor's modificationStamp and offset
3. we are invoking a non-AWT thread that should handle the completions fetching and rendering
4. we get the second "empty" event which modifies the editor's modificationStamp and offset - this event is blocked as we don't get completions for such a change
5. we are about to render the completions, but the modificationStamp and the offset are not equal to the new editor values, and therefore we are not rendering the completion.

When we are executing a non-AWT thread, so the order is inconsistent.

**The solution:**
Executing an AWT thread. this is the new order of events:
1. we get the first "\n" event
2. we are invoking an AWT thread that should handle the completions fetching and rendering, and then waiting
3. we get the second "empty" event which modifies the editor's modificationStamp and offset - this event is blocked as we don't get completions for such a change
4. the AWT thread starts running
5. we save the editor's modificationStamp and offset (now they are updated since the second event already fired)
6. we make sure that the editor values are equal
7. we are rendering the completion

before:

https://user-images.githubusercontent.com/10290661/209946430-74da6378-39bd-4d97-b41c-862c8154a175.mp4

after:

https://user-images.githubusercontent.com/10290661/209946405-20675fdf-784d-4dd6-a80c-8d92b19f6ba2.mp4

### DIFF
--- a/src/main/java/com/tabnine/inline/InlineCompletionHandler.java
+++ b/src/main/java/com/tabnine/inline/InlineCompletionHandler.java
@@ -1,7 +1,6 @@
 package com.tabnine.inline;
 
-import static com.tabnine.general.Utils.executeThread;
-import static com.tabnine.general.Utils.executeUIThreadWithDelay;
+import static com.tabnine.general.Utils.*;
 import static com.tabnine.prediction.CompletionFacade.getFilename;
 
 import com.intellij.openapi.application.ApplicationManager;
@@ -10,7 +9,6 @@ import com.intellij.openapi.editor.Document;
 import com.intellij.openapi.editor.Editor;
 import com.intellij.openapi.fileEditor.FileDocumentManager;
 import com.intellij.util.ObjectUtils;
-import com.intellij.util.concurrency.AppExecutorUtil;
 import com.tabnine.balloon.FirstSuggestionHintTooltip;
 import com.tabnine.binary.BinaryRequestFacade;
 import com.tabnine.binary.requests.autocomplete.AutocompleteResponse;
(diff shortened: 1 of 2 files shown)


src/main/java/com/tabnine/inline/InlineCompletionHandler.java
```diff
@@ -71,13 +65,45 @@ public void retrieveAndShowCompletion(
...
+              handleNewCompletions(
+                  editor,
+                  tabSize,
+                  updatedOffset,
+                  editor.getDocument().getModificationStamp(),
+                  completionAdjustment);
+            });
+  }
+
+  private void handleCachedCompletions(
```
yonip23, line 89: rename `handle` to `render` or something - "handle" is too generic imo

src/main/java/com/tabnine/inline/InlineCompletionHandler.java
```diff
@@ -71,13 +65,45 @@ public void retrieveAndShowCompletion(
...
     }
 
+    ApplicationManager.getApplication()
+        .invokeLater(
+            () -> {
+              int updatedOffset =
+                  editor.getCaretModel().getOffset()
+                      + (ApplicationManager.getApplication().isUnitTestMode()
+                          ? userInput.length()
+                          : 0);
```
yonip23, line 79: extract to function maybe?

yonip23 (approved):
can we write a test to this?
if not, can you attach a screenshot of before/after or something?
other than that, lgtm 🦈

yairco1990:
> 

unfortunately can't test it since `invokeLater` is mocked in the unit tests to be invoked immediately

dimacodota (approved):
LGTM

PR #450, opened 2022-12-29
DEV2-2042: refetch suggestions after debounce

yairco1990:
If debounce is enabled, re-fetch completions after debounce time has passed to get better completions.
In the video, we can see that after the debounce time we get new completions that we didn't get before.

https://user-images.githubusercontent.com/10290661/209967128-80496c60-74b3-408f-af28-1f9fdd97a705.mp4

### DIFF
--- a/src/main/java/com/tabnine/inline/InlineCompletionHandler.java
+++ b/src/main/java/com/tabnine/inline/InlineCompletionHandler.java
@@ -121,21 +121,28 @@ public class InlineCompletionHandler {
                 return;
               }
 
-              lastDebounceRenderTask =
-                  executeThread(
-                      () -> {
-                        List<TabNineCompletion> afterDebounceCompletions =
-                            retrieveInlineCompletion(editor, offset, tabSize, completionAdjustment);
-                        rerenderCompletion(
-                            editor,
-                            afterDebounceCompletions,
-                            offset,
-                            modificationStamp,
-                            completionAdjustment);
-                      },
-                      debounceTime,
-                      TimeUnit.MILLISECONDS);
+              refetchCompletionsAfterDebounce(
+                  editor, tabSize, offset, modificationStamp, completionAdjustment, debounceTime);
             });
+  }
+
+  private void refetchCompletionsAfterDebounce(
+      @NotNull Editor editor,
+      Integer tabSize,
+      int offset,
+      long modificationStamp,
+      @NotNull CompletionAdjustment completionAdjustment,
+      long debounceTime) {
+    lastDebounceRenderTask =
+        executeThread(
+            () -> {
+              List<TabNineCompletion> completions =
+                  retrieveInlineCompletion(editor, offset, tabSize, completionAdjustment);
+              rerenderCompletion(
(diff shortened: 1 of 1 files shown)


src/main/java/com/tabnine/inline/InlineCompletionHandler.java
```diff
@@ -125,14 +144,21 @@ private void rerenderCompletion(
...
       @NotNull CompletionAdjustment completionAdjustment) {
-    if (shouldCancelRendering(editor, modificationStamp, offset)) {
-      return;
-    }
-    if (shouldRemovePopupCompletions(completionAdjustment)) {
-      completions.removeIf(completion -> !completion.isSnippet());
-    }
-    showInlineCompletion(
-        editor, completions, offset, (completion) -> afterCompletionShown(completion, editor));
+    ApplicationManager.getApplication()
```
yonip23, line 147: doesn't this affect completions rendering? you wait for another AWT loop before rendering

yonip23 (commented):
please add a test

src/main/java/com/tabnine/inline/InlineCompletionHandler.java
yairco1990: `executeUIThreadWithDelay` already does that by scheduling a non-UI thread that in his turn, will invoke a UI thread.
I changed the behaviour to schedule a non-UI thread and then wrap the rendering function with a UI thread.
this is the flow:
1. document change event occurred
2. invoking a UI thread which will be invoked after all document changed events are fired (can be more than one for a single change)
3. invoking a non-UI thread for fetching completions
4. if debounce time is 0, execute the rendering function, which is wrapped by a UI thread. if denounce time is greater than 0 - schedule a non-UI thread that will invoke the rendering function.

I think it's safer to wrap the rendering method with a UI thread from the inside rather than manage it from the outside, which can lead to rendering in the non-UI thread

yonip23: I'm worried that since you're waiting a total of 2 UI event loops, you'll be messing with the user's actions in a buggy manner. For example, what happens if the user types / switches tab / etc.  between these 2 loop iterations you're waiting? it can lead to some messy edge cases IMO

yairco1990: you're right, and this is exactly the reason why we add this:

https://github.com/codota/tabnine-intellij/pull/450/files#diff-d6849f3d63babae30d241aa3112fdab621a3a77b2256ac66b4d51f6e87f02c8fR164

this is a validation of the offset and the modificationStamp between the first loop and the second loop so we ensure there is no change between them

yairco1990: anyway, you must invoke a new thread to fetch the completions, and then invoke a new UI thread for rendering, so there is no way to sneak out of that

yonip23 (approved):
lgtm

dimacodota (approved):
ךעאצ

PR #459, opened 2023-01-16
DEV2-2191 don't throw exception on index out of bounds

yonip23:


yairco1990:
can't we only check when we get the event if the offset is out of the document range and if so, ignore it?

yonip23:
> can't we only check when we get the event if the offset is out of the document range and if so, ignore it?

this will yield the same result as this pr, right? if you're in an invalid position you won't get completions.

i prefer to wrap it all with try-catch because all the other 3 lines in these functions are dangerous as well, so instead of understanding each edge case and writing 3 `if`s in both functions, I think its more robust to just catch any exception because all the handling is the same - just skip this completion.
wdyt?

dimacodota (approved):
lgtm

yairco1990 (approved):
lgtm

PR #466, opened 2023-01-31
DEV2-2022 add suggestion dropped event

yonip23:


src/main/java/com/tabnine/general/CompletionsEventSender.kt
```diff
@@ -23,12 +27,13 @@ class CompletionsEventSender(private val binaryRequestFacade: BinaryRequestFacad
         sendEventAsync(event)
     }
 
-    fun sendCancelSuggestionTrigger() {
-        val event = EventRequest("cancel-suggestion-trigger", mapOf())
+    fun sendSuggestionDropped(netLength: Int, filename: String?, metadata: CompletionMetadata?) {
+        val event = SuggestionDroppedRequest(netLength, filename, metadata)
+        System.err.println(event)
```
yonip23, line 32: remove

src/main/java/com/tabnine/inline/listeners/InlineFocusListener.kt
```diff
@@ -16,6 +15,6 @@ class InlineFocusListener(private val completionPreview: CompletionPreview) : Fo
 
     override fun focusGained(editor: Editor) {}
     override fun focusLost(editor: Editor) {
-        Disposer.dispose(completionPreview)
+//        Disposer.dispose(completionPreview)
```
yonip23, line 18: revert

src/main/java/com/tabnine/inline/InlineCompletionHandler.java
```diff
@@ -69,6 +75,16 @@ public void retrieveAndShowCompletion(
       return;
     }
 
+    if (lastShownSuggestion != null) {
+      // this means that the user did not type as suggested -
+      // we couldn't find completions in the cache, but there was a suggestion rendered
+      SuggestionDroppedReason reason =
+          completionAdjustment instanceof LookAheadCompletionAdjustment
+              ? SuggestionDroppedReason.ScrollLookAhead
+              : SuggestionDroppedReason.UserNotTypedAsSuggested;
```
yonip23, line 84: @yairco1990 this is how i differentiate `ScrollLookAhead` from `UserNotTypedAsSuggested` - wdyt?

src/main/java/com/tabnine/inline/CompletionPreview.java
```diff
@@ -97,6 +97,18 @@ public void togglePreview(CompletionOrder order) {
     completionsEventSender.sendToggleInlineSuggestionEvent(order, currentIndex);
   }
 
+  @Nullable
+  public static TabNineCompletion getCurrentCompletion(Editor editor) {
```
yairco1990, line 101: I think it will be better to move this function up to be near the other static methods

yairco1990 (approved):
great job!
I think we should consider creating a function (let's call it onSuggestionDropped) that gets the last completion and other params (such as reason, log message, etc.) and decide what to do with it (sending the drooped request if a suggestion exists for example). it seems to be a lot of logic around this event in almost every related code.
wdyt?

anyway, lgtm

src/main/java/com/tabnine/inline/InlineCompletionHandler.java
yairco1990: lgtm

yonip23:
note: merge this pr only after a matching binary version is released (with the `reason` field)

yairco1990 (approved):
love this!!

dimacodota (approved):
lgtm

PR #479, opened 2023-03-06
add default url

yanir-codota:


src/main/java/com/tabnine/userSettings/AppSettingsState.kt
```diff
@@ -26,7 +26,7 @@ class AppSettingsState : PersistentStateComponent<AppSettingsState?> {
     var debounceTime: Long = 0
     var autoImportEnabled: Boolean = true
     var binariesFolderOverride: String = ""
-    var cloud2Url: String = ""
+    var cloud2Url: String = "https://tabnine.sec.samsung.net"
```
assapir, line 29: Maybe pull it from somewhere?

PR #480, opened 2023-03-06
DEV2-2244 inject enterprise url from updater plugin

yonip23:
I tested it on ij 2022 and it works.
I don't think there's need to test on previous versions, as the `PropertiesComponent` is a stable API throughout the years since 2019, and I tested this API while testing the updater plugin.

### DIFF
--- a/src/main/java/com/tabnine/general/StaticConfig.java
+++ b/src/main/java/com/tabnine/general/StaticConfig.java
@@ -32,11 +32,6 @@ public class StaticConfig {
   public static final String TABNINE_ENTERPRISE_URL_PROPERTIES_KEY =
       TABNINE_PLUGIN_ID_RAW + ".enterpriseUrl";
 
-  // if the self-hosted updater injected the value, it'll be found here
-  @Nullable
-  public static final String TABNINE_ENTERPRISE_URL_DEFAULT_VALUE =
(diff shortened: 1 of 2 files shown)


src/main/java/com/tabnine/general/StaticConfig.java
```diff
@@ -22,11 +23,20 @@
...
 import javax.swing.*;
 import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
 
 public class StaticConfig {
   // Must be identical to what is written under <id>com.tabnine.TabNine</id> in plugin.xml !!!
   public static final String TABNINE_PLUGIN_ID_RAW =
       Config.IS_ON_PREM ? "com.tabnine.TabNine-Enterprise" : "com.tabnine.TabNine";
+  public static final String TABNINE_ENTERPRISE_URL_PROPERTIES_KEY =
+      TABNINE_PLUGIN_ID_RAW + ".enterpriseUrl";
```
yanir-codota, line 33: is this an entry in the settings?

yonip23: yes: https://github.com/tabnine/tabnine-jb-self-hosted-updater/blob/df93ee214a7d17e5fdb04d5bd779cf11ee002c9f/src/main/kotlin/com/tabnine/Consts.kt#L7

PR #483, opened 2023-03-08
DEV2-2435 add health check

yonip23:
previously, the update loop was initialized in the constructor of `BinaryStateServiec`.
the issue is that in onprem, we never called this service, so the constructor was never called.

i changed it so that the loop init is explicit and will be called within the `Initialize` extension point.


yairco1990 (approved):
lgtm

avnerbarr:
Lgtm

yanir-codota (approved):
LGTM

PR #488, opened 2023-03-23
DEV2-2518 add manual trigger inline completion action

yonip23:
users should define the shortcut for this action manually - no default shortcut is provided.
image


dimacodota (approved):
lgtm

PR #489, opened 2023-03-23
DEV2-2518 fix publish jobs

yonip23:
### Issue
the output of `./gradlew -q currentVersion` is like so (idk when it was changed):
```
➜  ./gradlew -q currentVersion
channelName: alpha
0.8.13
```
you can see here that the job thinks that `${{ steps.get_build_version.outputs.build_version }}` == `channelName: alpha`:
image

### Fix
add `| tail -n1` to the end of `./gradlew -q currentVersion`
this takes the last line of the output.
Its also idempotent, i.e. if there's only one line, it'll take this line.
You can see that if I run `tail -n1` twice it still outputs the correct version:
```
➜  CURRENT_VERSION=$(./gradlew -q currentVersion | tail -n1)
➜  echo $CURRENT_VERSION | tail -n1
0.8.13
```

### DIFF
--- a/.github/workflows/alpha.yml
+++ b/.github/workflows/alpha.yml
@@ -24,7 +24,7 @@ jobs:
       - name: Get Version
         id: get_build_version
         run: |
-          CURRENT_VERSION=$(./gradlew -q currentVersion)
+          CURRENT_VERSION=$(./gradlew -q currentVersion | tail -n1)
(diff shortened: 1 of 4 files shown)


amircodota:
Why do we need the channel name in the output? Wasn't it just a debug message that was accidentally left there?

amirbilu (approved):

yonip23:
@amircodota is fixing in another pr

PR #492, opened 2023-03-30
DEV2-2537: IJ plugin should install the binaries from the self-hosted updsate service

assapir:
Tested locally: 
Was able to get the version, download and run TabNine
```
00:00:00 /home/assaf/.tabnine/4.4.277/x86_64-unknown-linux-musl/TabNine --client IntelliJIDEA --no-lsp true --cloud2_url=http://localhost:3000 --client-metadata pluginVersion=1.0.0 clientIsUltimate=false clientChannel=onprem pluginUserId=ca109c96-95ad-49dc-ae13-a36185156caa debounceValue=300 clientVersion=2019.3 clientApiVersion=IC-193.5233.102 ide-restart-counter=0

```


src/main/java/com/tabnine/binary/fetch/BundleDownloader.java
```diff
@@ -23,6 +26,15 @@ public BundleDownloader(TempBundleValidator validator, GeneralDownloader downloa
 
   public Optional<BinaryVersion> downloadAndExtractBundle(String version) {
     String urlString = String.join("/", getBundleServerUrl(), version, TARGET_NAME, "TabNine.zip");
+
+    if (Config.IS_ON_PREM) {
+      if (!StaticConfig.getTabnineEnterpriseHost().isPresent()) {
```
dimacodota, line 31: you can chain `ifPresen(host ->  return host). orElseget( () -> {
Logger.getInstance(getClass()).warn("Not downloading bundle - on prem version but no host is setup!");
        return Optional.empty();}) `

assapir: I will move it anyway now

src/main/java/com/tabnine/general/StaticConfig.java
```diff
@@ -127,9 +137,9 @@ public static String getTabNineVersionUrl() {
   }
 
   @NotNull
-  public static String getTabNineBundleVersionUrl() {
+  public static String getTabNineBundleVersionUrl() throws NoValidBinaryToRunException {
     return Optional.ofNullable(System.getProperty(REMOTE_VERSION_URL_PROPERTY))
-        .orElse(getBundleServerUrl() + "/version");
+        .orElse(getBundleServerUrl().orElseThrow(NoValidBinaryToRunException::new) + "/version");
```
yonip23, line 142: im not sure about the error handling here - what happens if `NoValidBinaryToRunException` is thrown here? does the user see the ugly popup with this exception?
i think you should instead `warn` log this case, and return `Optional<String>` here, and the user should see the gray (inactive) status bar if the URL is not set (this is already implemented)

src/main/java/com/tabnine/general/StaticConfig.java
```diff
@@ -115,9 +115,19 @@ public static String getServerUrl() {
         .orElse("https://update.tabnine.com");
   }
 
-  public static String getBundleServerUrl() {
-    return Optional.ofNullable(System.getProperty(REMOTE_BASE_URL_PROPERTY))
-        .orElse("https://update.tabnine.com/bundles");
+  public static Optional<String> getBundleServerUrl() {
+    if (Config.IS_ON_PREM) {
+      if (!StaticConfig.getTabnineEnterpriseHost().isPresent()) {
```
yonip23, line 120: extract `StaticConfig.getTabnineEnterpriseHost()` to variable?

yonip23 (commented):
not sure about this (see comment)

src/test/java/com/tabnine/binary/BinaryVersionFetcherTests.java
```diff
@@ -30,10 +30,10 @@ public class BinaryVersionFetcherTests {
   @InjectMocks private BinaryVersionFetcher binaryVersionFetcher;
 
   @BeforeEach
-  public void setUp() throws BackingStoreException {
+  public void setUp() throws BackingStoreException, NoValidBinaryToRunException {
     Preferences preferences = Preferences.userNodeForPackage(BootstrapperSupport.class);
     preferences.clear();
-    when(binaryRemoteSource.fetchPreferredVersion(StaticConfig.getTabNineBundleVersionUrl()))
+    when(binaryRemoteSource.fetchPreferredVersion(anyString()))
```
yonip23, line 36: its better if you have the specific string to compare against - why did you change this? what broke the test?
same question for all the `anyString` changes

yonip23 (commented):
see comments
also, please provide some evidence of manual testing (either here / slack / doc / whatever)

src/test/java/com/tabnine/binary/BinaryVersionFetcherTests.java
assapir: because now I need to mock it inside

yonip23 (approved):
lgtm, well done for winning this battle with java 💪🏻

assapir:
> lgtm, well done for winning this battle with java 💪🏻

Thanks for the help ❤️

src/main/java/com/tabnine/general/StaticConfig.java
```diff
@@ -115,21 +114,27 @@ public static String getServerUrl() {
...
+      }
+      return Optional.of(String.join("/", tabnineEnterpriseHost.get(), "update", "bundles"));
+    }
 
-  @NotNull
-  public static String getTabNineVersionUrl() {
-    return Optional.ofNullable(System.getProperty(REMOTE_VERSION_URL_PROPERTY))
-        .orElse(getServerUrl() + "/version");
+    return Optional.of(
+        Optional.ofNullable(System.getProperty(REMOTE_BASE_URL_PROPERTY))
```
yanir-codota, line 128: what is this property? is it legit for non-onprem?

PR #496, opened 2023-04-03
DEV2-2535: Use IntelliJ Proxy settings

assapir:
By default, we are getting the proxy settings from IntelliJ and inheriting them to the binary.
This won't happen if the checkbox is disabled.
Was able to see the binary requests via `mitmproxy` when the setting was on.

Process env var in windows:
image

windows proxy:
image

### DIFF
--- a/src/main/java/com/tabnine/binary/BinaryProcessGateway.kt
+++ b/src/main/java/com/tabnine/binary/BinaryProcessGateway.kt
@@ -11,6 +11,7 @@ import java.io.InputStreamReader
 import java.net.Proxy
 import java.net.URL
 import java.nio.charset.StandardCharsets
+import java.util.Optional
 
 open class BinaryProcessGateway {
     private var process: Process? = null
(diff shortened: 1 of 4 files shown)


src/main/java/com/tabnine/binary/BinaryProcessGateway.kt
```diff
@@ -12,12 +18,40 @@ open class BinaryProcessGateway {
 
     @Throws(IOException::class)
     open fun init(command: List<String?>?) {
-        val createdProcess = ProcessBuilder(command).start()
+        val pb = ProcessBuilder(command)
```
yonip23, line 21: rename here and in the function argument to `processBuilder`

src/main/java/com/tabnine/binary/BinaryProcessGateway.kt
```diff
@@ -12,12 +18,40 @@ open class BinaryProcessGateway {
...
+    private fun setProxyEnvironmentVariables(pb: ProcessBuilder) {
+        val env = pb.environment()
+        val serverUrl = StaticConfig.getBundleServerUrl()
+        if (serverUrl.isPresent) {
+            val serverUrlString = serverUrl.get()
+            val tabnineServerProxy = CommonProxy.getInstance().select(URL(serverUrlString)).firstOrNull()
+            if (tabnineServerProxy?.type() == Proxy.Type.DIRECT) {
+                env["NO_PROXY"] = serverUrlString
+                env["no_proxy"] = serverUrlString
+            }
```
yonip23, line 40: extract to function `setOurServerUrlProxy`

src/main/java/com/tabnine/binary/BinaryProcessGateway.kt
```diff
@@ -12,12 +18,40 @@ open class BinaryProcessGateway {
...
+            return
+        }
+        val proxyString = "${HttpConfigurable.getInstance().PROXY_HOST}:${HttpConfigurable.getInstance().PROXY_PORT}"
+
+        if (proxyString.isNotBlank()) {
+            env["HTTPS_PROXY"] = proxyString
+            env["https_proxy"] = proxyString
+            env["HTTP_PROXY"] = proxyString
+            env["http_proxy"] = proxyString
+        }
```
yonip23, line 52: extract to function `inheritIdeProxy`

src/main/java/com/tabnine/binary/BinaryProcessGateway.kt
amirbilu: this should be the host, right? not including protocol

PR #506, opened 2023-04-16
DEV2-2533: Add plugin update endpoint as custome repo

assapir:
Ask the user to enter Cloud2 Enterprise URL if needed (based on the deprecated updater plugin).

When a user is setting/changing the cloud2Url, we will add/remove the `updatePlugin.xml` endpoint from the custom repository list.
image
image
image


src/main/java/com/tabnine/Initializer.java
```diff
@@ -65,6 +72,26 @@ private void initialize() {
         CapabilitiesService.getInstance().init();
         TabnineUpdater.pollUpdates();
         PluginInstaller.addStateListener(instanceOfUninstallListener());
+      } else if (Config.IS_ON_PREM && !DIALOG_SHOWED.get()) {
```
yonip23, line 75: i think there's `getAndSet` or something on `AtomicBoolean`

src/main/java/com/tabnine/Initializer.java
```diff
@@ -65,6 +72,26 @@ private void initialize() {
...
+        DIALOG_SHOWED.set(true);
+        Optional<String> cloud2Url = StaticConfig.getTabnineEnterpriseHost();
+        if (cloud2Url.isPresent()) {
+          Logger.getInstance(getClass()).info(String.format("Tabnine Enterprise host is configured: %s", cloud2Url.get()));
+          SetCustomRepository(cloud2Url.get());
+        } else {
```
yonip23, line 81: the `else` clause here is not right - you can get here by having `DIALOG_SHOWED == true` and `Config.IS_ON_PREM == false` - i.e., the else here doesn't necessarily mean that you're onprem.

src/main/java/com/tabnine/general/Utils.java
```diff
@@ -126,4 +131,30 @@ public static boolean isUnitTestMode() {
...
   }
+
+  public static String trimEndSlashAndWhitespace(String text) {
+    if (text == null) {
+      return null;
+    }
+    return text.replaceAll("/\\s*$", "");
+  }
+
+  public static void SetCustomRepository(String cloud2Url) {
```
yonip23, line 142: 1. this is not c# here lol
2. rename argument
```suggestion
  public static void setCustomRepository(String url) {
```

src/main/java/com/tabnine/general/Utils.java
```diff
@@ -126,4 +131,30 @@ public static boolean isUnitTestMode() {
...
+    if (!cloud2Url.trim().isEmpty()) {
+      List<String> newPluginRepo = UpdateSettings.getInstance().getStoredPluginHosts();
```
yonip23, line 144: wrong variable name
```suggestion
      List<String> pluginsHosts = UpdateSettings.getInstance().getStoredPluginHosts();
```

src/main/java/com/tabnine/general/Utils.java
```diff
@@ -126,4 +131,30 @@ public static boolean isUnitTestMode() {
...
+      String newStore = String.format("%s/update/jetbrains/updatePlugins.xml", trimEndSlashAndWhitespace(cloud2Url));
+      newPluginRepo.add(newStore);
+      Logger.getInstance(Utils.class).debug(String.format("Added custom repository to %s", newStore));
+      removeDuplicates(newPluginRepo);
+    }
+  }
+
+  public static void RemoveCustomRepository(String oldCloud2Url) {
```
yonip23, line 152: same comments
```suggestion
  public static void removeCustomRepository(String oldUrl) {
```

src/main/java/com/tabnine/general/Utils.java
```diff
@@ -126,4 +131,30 @@ public static boolean isUnitTestMode() {
...
+    if (!oldCloud2Url.trim().isEmpty()) {
+      List<String> hostsList = UpdateSettings.getInstance().getStoredPluginHosts();
+      String oldPluginRepo = String.format("%s/update/jetbrains/updatePlugins.xml", trimEndSlashAndWhitespace(oldCloud2Url));
```
yonip23, line 155: same comment
```suggestion
      String pluginHosts = String.format("%s/update/jetbrains/updatePlugins.xml", trimEndSlashAndWhitespace(oldCloud2Url));
```

src/main/java/com/tabnine/userSettings/AppSettingsState.kt
```diff
@@ -27,6 +29,11 @@ class AppSettingsState : PersistentStateComponent<AppSettingsState?> {
     var autoImportEnabled: Boolean = true
     var binariesFolderOverride: String = ""
     var cloud2Url: String = ""
+        set(value) {
```
yonip23, line 32: im not in favor of performing these operations here, please move this logic to [here](https://github.com/codota/tabnine-intellij/blob/45de31701d23f424aa360d06aa0dcf73fcf9fc23/src/main/java/com/tabnine/userSettings/AppSettingsConfigurable.kt#L47) or some other method(s) in this file

src/main/java/com/tabnine/userSettings/AppSettingsState.kt
```diff
@@ -27,6 +29,11 @@ class AppSettingsState : PersistentStateComponent<AppSettingsState?> {
...
+            SetCustomRepository(value)
+            RemoveCustomRepository(field)
```
yonip23, line 34: i'd create a separate util function `replaceCustomRepository(String old, String new)` - that will call `UpdateSettings.getInstance().getStoredPluginHosts()` only once.
I don't know how this API behaves when called twice after being updated, and even if it works as expected, its more reasonable to perform both changes (delete and add) using the same list obtained by this API.

yonip23 (changes requested):
im requesting changes mainly for the comment around `DIALOG_SHOWED` and the one regarding the `cloud2Url`'s `set` function

src/main/java/com/tabnine/Initializer.java
assapir: that an `else` for `if (cloud2Url.isPresent())`, not the one in the upper scope

src/main/java/com/tabnine/userSettings/AppSettingsState.kt
assapir: but... that what you actually have setters for...

yonip23: Setters are for lightweight operations of coupled states, and I think its extra clear in Kotlin where the syntactic sugaring for setters is literally the `=` operation - you can't expect heavy operations to be performed on assignment, right?

but ` UpdateSettings.getInstance().storedPluginHosts` is a more serious operation - it includes `ServiceManager.getService`, which can be IDE/version specific, and it includes multiple file reads which can not only fail but slow down the process or something.

Invoking all the above on a setter can be misleading, especially when it's being called not only from the UI component.

src/main/java/com/tabnine/Initializer.kt
```diff
@@ -0,0 +1,88 @@
...
+                }
+            }
+        }
+    }
+
+    companion object {
+        private val connectionLostNotificationHandler = ConnectionLostNotificationHandler()
+        private val dialogShown = AtomicBoolean(false)
+        private val initialized = AtomicBoolean(false)
+    }
```
yonip23, line 87: there's no reason to put them in a companion object, they can be regular fields. that's part of the conversion mishaps I was talking about lol

src/main/java/com/tabnine/Initializer.kt
```diff
@@ -0,0 +1,88 @@
...
+                        .warn(
+                            "Tabnine Enterprise host is not configured, showing some nice dialog"
+                        )
+                    ApplicationManager.getApplication().invokeLater(
+                        Runnable {
+                            val dialog = TabnineEnterpriseUrlDialogWrapper(null)
+                            if (dialog.showAndGet()) {
+                                val url = dialog.inputData
+                                instance.cloud2Url = url
+                                Utils.setCustomRepository(url)
```
yonip23, line 73: if you're calling `replaceCustomRepository` inside the setter, why do you call `setCustomRepository` here? is it possible that you got misled by the setter thingy...?

src/main/java/com/tabnine/Initializer.kt
```diff
@@ -0,0 +1,88 @@
...
+    private var binaryPromotionStatusBarLifecycle: BinaryPromotionStatusBarLifecycle? = null
+    override fun preload(indicator: ProgressIndicator) {
+        initialize()
+    }
+
+    override fun runActivity(project: Project) {
+        initialize()
+    }
+
+    private fun initialize() {
```
yonip23, line 38: 1. i think the `dialogShown` boolean is redundant, I don't see any reason why not to use the `initialized` atomic boolean for this - i.e. if not initizlied && on prem => show dialog
2. the flow of this function would be much clearer IMO like so:
```kotlin
if (!shouldInitialize) return
// ... rest of the code
if (Config.IS_ON_PREM) {
  requireUserCloud2Url();
} else {
  initializeNonOnPremListeners();
}
```
where `requireUserCloud2Url` will contain all the `if` and `else` clauses of the `dialogShwon` related code (i.e. show the dialog if needed and set the custom repo URL)

and `initializeNonOnPremListeners` will contain all the code under the `!Config.IS_ON_PREM` condition in the current code

src/main/java/com/tabnine/Initializer.kt
```diff
@@ -0,0 +1,88 @@
...
+            Logger.getInstance(javaClass)
+                .info(
+                    "Initializing for ${Config.CHANNEL}, plugin id = ${StaticConfig.TABNINE_PLUGIN_ID_RAW}"
+                )
+            connectionLostNotificationHandler.startConnectionLostListener()
+            ServiceManager.getService(BinaryStateService::class.java).startUpdateLoop()
+            initTabnineLogger()
+            if (!Config.IS_ON_PREM) {
+                binaryNotificationsLifecycle = DependencyContainer.instanceOfBinaryNotifications()
+                binaryPromotionStatusBarLifecycle = DependencyContainer.instanceOfBinaryPromotionStatusBar()
```
yonip23, line 50: i think its unnecessary to initialize these fields here and not in the constructor (just like `connectionLostNotificationHandler`) - I think it's a small mistake that stayed with us for a long time. When you convert it to Kotlin, this mistake becomes clearer.

I'd change these fields to be initialized in the beginning and ensure everything is working by building a local plugin for production and ensuring you receive notifications and such.

yonip23 (commented):
Well done for the effort to convert files to Kotlin 👏🏻 
I still have some comments so not approving yet, but you're doing a great hob non the less 💪🏻

src/main/java/com/tabnine/Initializer.kt
assapir: nope, the atomic bool need to be static (or class field, not instance) because it is called twice (not sure why)

src/main/java/com/tabnine/userSettings/AppSettingsState.kt
assapir: The problem with where you suggested - it's will call only when it changed via the UI, but I want the effect to be called where ever it changed, not matter if via code or via UI.

yonip23 (approved):
lgtm

PR #509, opened 2023-04-17
DEV2-2533: Download and install enterprise plugin

assapir:
What have we done here:
1. Split the solution into three components:
    1. `Common` - Where most of the code resides. It has `plugin.xml` only for tests to work
    2. `Tabnine`  -  The public plugin. Have all the channels and is published by the CI to the public marketplace. Has a `plugin.xml` file with the public plugin ID.
    3. `TabnineSelfHosted` - The self-hosted plugin. Should never be published to the marketplace, and always use the `self_hosted` channel (until fully removed in the next PRs). Has the `plugin.xml` with the `-Enterprise` suffix.
2. To run any `gradle` task, one has a couple of options:
     1. Run the on all components: `./gradlew <task>`, e.g. `./gradlew build` will build all the components.
     2. Run only for one component: `./gradlew :<component>:<task>`, e.g., `./gradlew :TabnineSelfHosted:build` will only build the self-hosted plugin.
3. CI:  It should be the same.as used to be.
    1. Add an ID verification test also for the self-hosted plugin.
    2. Build both plugins to make sure it isn't broken.
    3. run `runPluginVerifier` for both plugins.
5. CD: 
    1. `Alpha Release` will build & release **only** the public plugin.
    2. `Production Release With Version` will build and release **both** versions, where the self-hosted one will go to GCP to be collected by the [self-hosted-update-service](https://github.com/codota/self-hosted-update-service/) when needed.


src/main/java/com/tabnine/general/StaticConfig.java
```diff
@@ -16,227 +16,230 @@
...
-
-  public static Optional<String> getBundleServerUrl() {
-    if (Config.IS_SELF_HOSTED) {
-      Optional<String> tabnineEnterpriseHost = StaticConfig.getTabnineEnterpriseHost();
-      if (!tabnineEnterpriseHost.isPresent()) {
-        Logger.getInstance(StaticConfig.class).warn("On prem version but server url not set");
-        return Optional.empty();
-      }
-      return Optional.of(String.join("/", tabnineEnterpriseHost.get(), "update", "bundles"));
+    // Must be identical to what is written under <id>com.tabnine.TabNine</id> in plugin.xml !!!
```
yonip23, line 29: please revert all the format changes here, its confusing to understand the actual diff

src/main/java/com/tabnine/lifecycle/EnterprisePluginInstaller.kt
```diff
@@ -0,0 +1,99 @@
...
+import com.intellij.util.xmlb.annotations.Attribute
+import com.intellij.util.xmlb.annotations.Tag
+import com.tabnine.dialogs.Dialogs
+import com.tabnine.general.StaticConfig
+import com.tabnine.general.StaticConfig.TABNINE_ENTERPRISE_ID_RAW
+import com.tabnine.general.Utils
+import java.net.URL
+import java.util.concurrent.locks.ReentrantLock
+
+data class Plugin(
```
yonip23, line 23:
```suggestion
data class TabninePluginDescriptor(
```

src/main/java/com/tabnine/lifecycle/EnterprisePluginInstaller.kt
```diff
@@ -0,0 +1,99 @@
...
+                        downloadAndInstall(indicator, plugin)
+                    }
+                } catch (e: Throwable) {
+                    Logger.getInstance(javaClass).warn("Failed to install Tabnine Enterprise plugin", e)
+                }
+            }
+        })
+    }
+
+    private fun getPluginDescriptor(host: String): Plugin? {
```
yonip23, line 53:
```suggestion
    private fun getTabninePluginDescriptor(host: String): Plugin? {
```
just so it'll be clear that its from our repo and not some GA marketplace

src/main/java/com/tabnine/lifecycle/EnterprisePluginInstaller.kt
```diff
@@ -0,0 +1,99 @@
...
+        val element = JDOMUtil.load(URL(url.get()))
+        return XmlSerializer.deserialize(element.getChild("plugin"), Plugin::class.java)
+    }
+
+    private fun downloadAndInstall(
+        indicator: ProgressIndicator,
+        plugin: Plugin,
+    ) {
+        val newVersion = plugin.parsedVersion ?: return
+        val downloadUrl = plugin.url ?: return
```
yonip23, line 64: please add logs for these cases

.github/workflows/production-new-version.yml
```diff
@@ -149,6 +149,9 @@ jobs:
           key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle') }}
           restore-keys: ${{ runner.os }}-gradle
 
+      - name: Change plugin id
```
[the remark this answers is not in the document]
dimacodota: This will be the same for vscode, it will be a different extension with its package.json

src/main/java/com/tabnine/general/StaticConfig.java
assapir: You can hide whitespaces diff... but will revert

.github/workflows/ci.yml
```diff
@@ -54,11 +54,14 @@ jobs:
           mkdir latest && tar -xvf latest.tar.gz -C latest --strip-components 1
           ls -al latest
 
-      - name: Build the plugin
+      - name: Build the plugins
         run: ./gradlew buildPlugin
 
-      - name: Validate the plugin
-        run: ./gradlew runPluginVerifier -PlocalPaths="$(pwd)/latest"
+      - name: Validate the regular plugin
```
yonip23, line 60:
```suggestion
      - name: Validate the public plugin
```

.github/workflows/ci.yml
```diff
@@ -54,11 +54,14 @@ jobs:
           mkdir latest && tar -xvf latest.tar.gz -C latest --strip-components 1
           ls -al latest
 
-      - name: Build the plugin
+      - name: Build the plugins
```
yonip23, line 57: i'd separate to 2 jobs so that they can run in parallel

.github/workflows/compatibility-eap.yml
```diff
@@ -22,7 +22,7 @@ jobs:
           key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle') }}
           restore-keys: ${{ runner.os }}-gradle
 
-      - name: Build the plugin
+      - name: Build the plugins
```
yonip23, line 25: here as well - 2 jobs is better imo

.github/workflows/production-new-version.yml
```diff
@@ -150,11 +150,11 @@ jobs:
           restore-keys: ${{ runner.os }}-gradle
 
       - name: Build the plugin
```
yonip23, line 152:
```suggestion
      - name: Build the self hosted plugin
```

.github/workflows/production-new-version.yml
```diff
@@ -150,11 +150,11 @@ jobs:
...
-        run: ./gradlew buildPlugin -PexternalVersion=${{ needs.set-version.outputs.build_version }} -Pchannel=self_hosted
+        run: ./gradlew :TabnineSelfHosted:buildPlugin -PexternalVersion=${{ needs.set-version.outputs.build_version }} -Pchannel=self_hosted
 
       - name: Get Plugin Path
```
yonip23, line 155:
```suggestion
      - name: Upload self hosted plugin to GitHub
```
?

yonip23 (commented):
other than the (minor) comments added here, I have some more general / not file-specific comments:
1. please document the changes in good detail - including:
    - CI - how it runs, and how to release each plugin
    - package description, which package contains what
    - local development - what a dev needs to do in order to work with the plugin (build, runIde, tests etc.)
2. please make sure you test these changes thoroughly - this is a breaking change and we need to release it gradually carefully.
I'd suggest releasing to alpha from this branch, and:
    - make sure there's no regression
    - maybe have an online session with 1 or 2 devs to make sure they're able to setup and run the plugin locally (tests etc.)
3. the resources are not well organized -
    1. remove `icons` folder from the test resources in `Common`
    2. move all basic (common) icons to the `Common` project
    3. extract plugin-specific icons to their respective projects

.github/workflows/compatibility-eap.yml
assapir: Why. most of the time is building `common`. this way it builds only once.

.github/workflows/production-new-version.yml
assapir: what, nom it's not what this step is doing.

.github/workflows/compatibility-eap.yml
yonip23: mmm but if the `Tabnine` build breaks you don't get to see if the `SelfHosted` build passes or not.
I don't really care either way, need to work with the CI a bit to get a sense of what's working best.

yonip23 (approved):
lgtm

TabnineSelfHosted/build.gradle
```diff
@@ -0,0 +1,99 @@
...
+}
+
+repositories {
+    mavenCentral()
+}
+
+apply plugin: "org.jlleitschuh.gradle.ktlint-idea"
+
+dependencies {
+    implementation(project(":Common"))
```
amircodota, line 36: maybe we can clean some of these dependencies here?

I mean - it is probably included in Common, right? are we sure we also need it here?

TabnineSelfHosted/build.gradle
```diff
@@ -0,0 +1,99 @@
...
+    testRuntimeOnly 'org.junit.vintage:junit-vintage-engine:5.9.0'
+}
+
+intellij {
+    version = '2019.3'
+    type = 'IC'
+    updateSinceUntilBuild = false
+}
+
+def PRODUCTION_CHANNEL = null
```
amircodota, line 56: Is this stuff needed for self hosted?

amircodota (approved):
Had some minor comments about gradle, but overall LGTM.

I think we should add som documentation - probably in the README.

Also, can we try to make it a little more foolproof? I mean, I think we can pretty easily publish the wrong thing to the wrong place. The SelfHosted to the public one or vice versa?

PR #515, opened 2023-04-24
DEV2-2621 - do not set the https_proxy env because jb setting does not expose the…

dimacodota:
do not set the https_proxy env because jb setting does not expose the schema


dimacodota:
tested as described here https://tabninedesk.zendesk.com/agent/tickets/22837

PR #525, opened 2023-05-04
DEV2-2560: Fix publish with wrong name

assapir:
now it will always produce the same plugin name even if the project name is different


yairco1990 (approved):

amircodota (approved):
LGTM

PR #526, opened 2023-05-07
DEV2-2704: Sign in / sign out / get state from JB plugin

assapir:
When the user is not logged in: 
image
image
image

When the user is logged in, and host is set: 
image
image
image

Logged in without server URL:
image


TabnineSelfHosted/src/main/java/com/tabnineSelfHosted/statusBar/SelfHostedStatusBarActions.kt
```diff
@@ -0,0 +1,66 @@
...
+        return DumbAwareAction.create(OPEN_TABNINE_SETTINGS_TEXT) {
+            Logger.getInstance(javaClass).info("Opening Tabnine settings")
+            ShowSettingsUtil.getInstance()
+                .editConfigurable(project, AppSettingsConfigurable())
+        }
+    }
+
+    private fun createLoginAction(): DumbAwareAction {
+        return DumbAwareAction.create(LOGIN_TEXT) {
+            Logger.getInstance(javaClass).info("Logging in")
```
yairco1990, line 51: why not just "Login" and "Logout"?

assapir: Product ask

Common/src/main/java/com/tabnineCommon/binary/requests/config/StateResponse.kt
```diff
@@ -12,7 +12,11 @@ data class StateResponse(
     @SerializedName("installation_time")
     val installationTime: String? = null,
     @SerializedName("cloud_connection_health_status")
-    val cloudConnectionHealthStatus: CloudConnectionHealthStatus = CloudConnectionHealthStatus.Ok
+    val cloudConnectionHealthStatus: CloudConnectionHealthStatus = CloudConnectionHealthStatus.Ok,
+    @SerializedName("is_logged_in")
+    val isLoggedIn: Boolean? = null,
+    @SerializedName("user_name")
+    var userName: String? = null,
```
yonip23, line 19: isn't the `isLoggedIn` boolean redundant? isn't it equivalent to `userName != null`?

Common/src/main/resources/META-INF/common-plugin.xml
```diff
@@ -105,7 +105,7 @@
 
     <change-notes>Tabnine is always evolving. Most updates are in the binary sidecar.</change-notes>
 
-    <idea-version since-build="193"/>
+    <idea-version since-build="193.5233.102"/>
```
yonip23, line 108: please verify (and share with us) in mixpanel that we're not ditching too many users here.
should be fine, but I think you should at least fetch some insights from mixpanel consult with your team lead or someone.

also, this can be extracted to a `gradle.properties` file I think, just so it won't repeat itself multiple times in this repo. but not critical for now.

TabnineSelfHosted/src/main/java/com/tabnineSelfHosted/statusBar/TabnineSelfHostedStatusBarWidget.kt
```diff
@@ -0,0 +1,111 @@
...
+                BinaryStateChangeNotifier.STATE_CHANGED_TOPIC,
+                BinaryStateChangeNotifier { (_, _, _, cloudConnectionHealthStatus1, _, userName): StateResponse ->
+                    cloudConnectionHealthStatus = cloudConnectionHealthStatus1
+                    username = userName
+                    update()
+                }
+            )
+    }
+
+    override fun getIcon(): Icon? {
```
yonip23, line 40: consider extracting some of the logic in the functions below to separate file/s for better readability of the files

yonip23 (commented):
overall nice work 👍🏻, see comments.
also, one general comment:
since we have the disappearing status bar issues, and regardless I think its a better UX, i'd add the login/logout operations as actions available from the ctrl+shift+A shortcut - example can be found here:
code: https://github.com/codota/tabnine-jb-self-hosted-updater/blob/main/src/main/kotlin/com/tabnine/ConfigureUrlAction.kt
XML configuration: https://github.com/codota/tabnine-jb-self-hosted-updater/blob/7739c84b905e0e341168c5d17837f62303438b3f/src/main/resources/META-INF/plugin.xml#L31

Common/src/main/java/com/tabnineCommon/binary/requests/config/StateResponse.kt
assapir: we get it anyway from the binary

Common/src/main/resources/META-INF/common-plugin.xml
assapir: ` 193.5662.53` is the first version of `2019.3` -> `2019.3.1`, according to https://www.jetbrains.com/idea/download/other.html. So I don't think it actually matter but not getting the  compile error

Common/src/main/java/com/tabnineCommon/binary/requests/config/StateResponse.kt
yonip23: so maybe worth removing from the binary as well🤷🏻‍♂️

yonip23 (approved):

amirbilu (approved):

PR #530, opened 2023-05-24
DEV2-2807: Explanatory IDE indications in status bar

assapir:


TabnineSelfHosted/src/main/java/com/tabnineSelfHosted/binary/lifecycle/UserInfoService.kt
```diff
@@ -0,0 +1,38 @@
...
+import com.tabnineSelfHosted.binary.requests.userInfo.UserInfoResponse
+import java.util.concurrent.TimeUnit
+import java.util.concurrent.atomic.AtomicBoolean
+
+class UserInfoService {
+    private val scheduler = AppExecutorUtil.getAppScheduledExecutorService()
+    private val messageBus: MessageBus = ApplicationManager.getApplication().messageBus
+    private val updateLoopStarted = AtomicBoolean(false)
+    private val binaryRequestFacade = DependencyContainer.instanceOfBinaryRequestFacade()
+    private var userInfoResponse: UserInfoResponse? = null
```
yonip23, line 17: why do you keep the user info response here if its already saved in `UserInfoService`?

TabnineSelfHosted/src/main/java/com/tabnineSelfHosted/statusBar/SelfHostedStatusBarActions.kt
```diff
@@ -10,27 +11,34 @@ import com.tabnineCommon.binary.requests.login.LoginRequest
...
-        )
+        if (AppSettingsState.instance.cloud2Url.isNotBlank()) {
+            actions.add(
+                if (isLoggedIn == null) {
+                    createGoToFAQAction()
+                } else if (isLoggedIn) {
+                    createLogoutAction()
+                } else {
+                    createLoginAction()
+                }
```
yonip23, line 39: i don't like this usage of `null`s - i'd create an enum `UserLoginStatus` with 3 options: `LoggedIn`, `LoggedOut`, and `NotAvailable` or something.

TabnineSelfHostedForMarketplace/src/main/java/com/tabnineSelfHosted/userSettings/AppSettingsState.kt
```diff
@@ -1,11 +1,14 @@
 package com.tabnineSelfHosted.userSettings
 
+import com.intellij.ide.util.PropertiesComponent
```
yonip23, line 3: same comments on this file as in the other settings state implementation

TabnineSelfHosted/src/main/java/com/tabnineSelfHosted/statusBar/TabnineSelfHostedStatusBarWidget.kt
```diff
@@ -14,38 +14,56 @@ import com.tabnineCommon.binary.requests.config.CloudConnectionHealthStatus
...
+                }
+            )
+
+        ApplicationManager.getApplication()
+            .messageBus
+            .connect(this)
+            .subscribe(
+                UserInfoChangeNotifier.USER_INFO_CHANGED_TOPIC,
+                UserInfoChangeNotifier { userInfoResponse ->
+                    userInfo = userInfoResponse
```
yonip23, line 49: I don't like that you're saving all the states coming from the notifiers (the `cloudConnectionHealthStatus` and the `userInfo` fields) - can't you use the `UserInfoService` for this? It already saves the last user info response.
The same goes for the `cloudConnectionHealthStatus` - I can't remember if it has a matching service similar to the `UserInfoService`, and if it not, it might be worth creating one.

going by your implementation is dictating a bad design, where every consumer of these states saves a copy of the same response.

yonip23 (changes requested):
overall great job, but there are some changes i'd like you to make before merging 🙏🏻

yonip23 (approved):
Lgtm :rocket:

amirbilu (approved):

PR #533, opened 2023-05-30
DEV2-2856 re-fetch suggestions when results are empty

yonip23:
using the existing debounce mechanism which fetches before debouncing and then re-fetches after, just applying the same logic with debounce time = 800ms when results are empty. 

you can see in the attached video that I did not type as suggested multiple times, and received completions each time.

https://github.com/codota/tabnine-intellij/assets/67855609/a539f6cb-1a06-4839-98b0-1b7244361cf0


dimacodota (approved):
LGTM

PR #543, opened 2023-06-22
DEV2-2972 Tabnine chat (alpha only)

yonip23:


Common/src/main/java/com/tabnineCommon/capabilities/CapabilitiesService.java
```diff
@@ -55,8 +60,6 @@ private void fetchCapabilitiesLoop() {
     Optional<Long> lastPid = Optional.empty();
 
     try {
-      Thread.sleep(INITIAL_DELAY_MS);
-
```
yonip23, line 59: @amircodota we (@yairco1990 and i) decided to remove it, as it caused the chat to have a delay of 5-10 seconds on start, and without it its much better (idk what kind of math converts 2s delay to almost 10s delay, but 🤷🏻‍♂️ )

wdyt? should we worry for bad side effects?

amircodota: I honestly do not remember. Sorry :-)

Tabnine/src/main/java/com/tabnine/chat/commandHandlers/InitHandler.kt
```diff
@@ -0,0 +1,40 @@
...
+    private val isDarkTheme: Boolean,
+    private val colors: MutableMap<String, String>,
+    private val fontSize: Int,
+)
+
+class InitHandler(gson: Gson) : ChatMessageHandler<Unit, InitPayload>(gson) {
+    override fun handle(payload: Unit?, project: Project): InitPayload {
+        val colorPalette = readColorPalette()
+        val isDarkTheme = EditorColorsManager.getInstance().isDarkEditor
+        val font = EditorColorsManager.getInstance().globalScheme.getFont(EditorFontType.PLAIN).size
```
yonip23, line 21: @yairco1990 note that this is representing the `Editor -> Font -> Size` setting property, so it should be fine, but it doesn't adapt to ctrl+mouse wheel font changes

Tabnine/src/main/java/com/tabnine/BrowserCreator.kt
```diff
@@ -0,0 +1,133 @@
...
+import com.tabnineCommon.general.StaticConfig
+import org.cef.browser.CefBrowser
+import org.cef.browser.CefFrame
+import org.cef.handler.CefLoadHandler
+import org.cef.network.CefRequest
+import java.awt.Toolkit
+import java.awt.datatransfer.StringSelection
+import java.nio.file.Paths
+
+object BrowserCreator {
```
dimacodota, line 22: this is chat specific impl, so id move to  a chat area

Tabnine/src/main/java/com/tabnine/chat/TabnineChatProjectManagerListener.kt
```diff
@@ -0,0 +1,95 @@
...
+import com.intellij.openapi.wm.ToolWindowAnchor
+import com.intellij.openapi.wm.ToolWindowManager
+import com.tabnine.BrowserCreator
+import com.tabnineCommon.capabilities.CapabilitiesService
+import com.tabnineCommon.capabilities.Capability
+import com.tabnineCommon.lifecycle.BinaryCapabilitiesChangeNotifier
+import java.util.concurrent.atomic.AtomicBoolean
+import javax.swing.JPanel
+
+class TabnineChatProjectManagerListener private constructor() : ProjectManagerListener, Disposable {
```
dimacodota, line 21: serviceFactotyCreatorBuilder 🤣

Tabnine/src/main/java/com/tabnine/chat/ChatBundleExtractor.kt
```diff
@@ -0,0 +1,49 @@
...
+
+private const val CHAT_BUNDLE_TAR_FILE = "chat-bundle.tar.gz"
+
+object ChatBundleExtractor {
+
+    @Throws(IllegalStateException::class)
+    fun extractBundle(destination: Path) {
+        val resource = javaClass.classLoader.getResourceAsStream(CHAT_BUNDLE_TAR_FILE)
+            ?: throw IllegalStateException("Unable to find resource $CHAT_BUNDLE_TAR_FILE")
+        untar(resource, destination.toFile())
```
dimacodota, line 21: why is it extracted at runtime and not  in the CI as we did in vscode? @yairco1990

yonip23: because `javaClass.classLoader.` doesn't have a f.s.-like API, you can't "get all resources under x", you need to find each resource by name.

And since these files include js map files and such, I decided that the best solution would be to bundle a single resource which is the tar, and simply untar it at runtime

amirbilu (approved):
LGTM by @dimacodota

PR #546, opened 2023-06-27
DEV2-2972 catch browser creation error gracefully

yonip23:
i moved the browser creation to earlier in the flow, before registering the tool window - in order to catch its creation failure beforehand

### DIFF
--- a/Tabnine/src/main/java/com/tabnine/BrowserCreator.kt
+++ b/Tabnine/src/main/java/com/tabnine/BrowserCreator.kt
@@ -1,0 +1,133 @@
+package com.tabnine
+
+import com.intellij.openapi.application.ApplicationManager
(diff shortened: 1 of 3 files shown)


Tabnine/src/main/java/com/tabnine/BrowserCreator.kt
```diff
@@ -0,0 +1,133 @@
...
+import com.tabnineCommon.general.StaticConfig
+import org.cef.browser.CefBrowser
+import org.cef.browser.CefFrame
+import org.cef.handler.CefLoadHandler
+import org.cef.network.CefRequest
+import java.awt.Toolkit
+import java.awt.datatransfer.StringSelection
+import java.nio.file.Paths
+
+object BrowserCreator {
```
yonip23, line 22: moved from the factory class

PR #547, opened 2023-06-27
DEV2-3023 use user avatar

dimacodota:


Common/src/main/java/com/tabnineCommon/binary/requests/config/StateResponse.kt
```diff
@@ -19,6 +19,8 @@ data class StateResponse(
     var userName: String? = null,
     @SerializedName("access_token")
     var accessToken: String? = null,
+    @SerializedName("user_avatar_url")
```
yonip23, line 22: Use camel case

yonip23 (approved):

PR #552, opened 2023-06-29
DEV2-2972 add "ask chat" action

yonip23:
image

note: this action doesn't have a default shortcut, so users should define one in Settings -> `Keymap -> "Ask Chat"`.
image


yairco1990 (approved):
looks great!
the only thing I would change is to make the icon aligned to the same line of the input :)

yonip23:
> looks great! the only thing I would change is to make the icon aligned to the same line of the input :)

I'm using `Messages.showInputDialog`, and that's where they place the icon. To change the icon location I need to create a custom dialog, which I don't think is worth the effort for now

dimacodota (approved):
LGTM

PR #558, opened 2023-07-03
DEV2-3062 Do not run binary versions 4.5.0 through 4.5.13

amircodota:
they have an isue updating in envs with custom certificates, so it is up to the plugin to download a newer version with a fix


dimacodota (approved):
lgtm

PR #559, opened 2023-07-05
DEV2-2972 add reload browser option

yonip23:
We get alpha users complaining from time to time that the UI doesn't render inside the browser, so I added this action in hopes that it'll be able to remediate the issue when it appears.


yairco1990 (approved):
:100:

dimacodota (approved):
lgtm

PR #564, opened 2023-07-09
DEV2-3077: Enable chat in IJ for self hosted

assapir:


Common/src/main/java/com/tabnineCommon/chat/TabnineChatProjectManagerListener.kt
```diff
@@ -36,6 +38,15 @@ class TabnineChatProjectManagerListener private constructor() : ProjectManagerLi
     fun start() {
         if (initialized.getAndSet(true)) return
 
+        if (StaticConfig.getTabnineEnterpriseHost().isPresent) {
```
yonip23, line 41: don't we have a `Channel.isOnprem` or something?

yonip23, line 41: don't we have a `Channel.isOnprem` or something?
the channel check is stronger because its determined in compile time, so i'd consider checking it in addition to the available host check

Common/src/main/java/com/tabnineCommon/chat/commandHandlers/InitHandler.kt
```diff
@@ -23,13 +25,21 @@ class InitHandler(gson: Gson) : ChatMessageHandler<Unit, InitPayload>(gson) {
...
     private fun isTelemetryEnabled(): Boolean {
         return CapabilitiesService.getInstance().isCapabilityEnabled(Capability.ALPHA)
     }
 
+    private fun getServerUrl(): String? {
+        return if (StaticConfig.getTabnineEnterpriseHost().isPresent) {
+            StaticConfig.getTabnineEnterpriseHost().get()
+        } else {
+            null
+        }
```
yonip23, line 40:
```suggestion
        return StaticConfig.getTabnineEnterpriseHost().orElse(null)
```

Common/src/main/java/com/tabnineCommon/chat/TabnineChatProjectManagerListener.kt
yonip23, line 41: also, not critical, but i'd consider creating a function that accepts a function: `invokeWhenChatEnabled`, that'll check if isOnprem or capability enabled. Then you use it like so:
```kotlin
invokeWhenChatEnabled {
  ProjectManager.getInstance().openProjects.forEach { projectOpened(it) }
  ApplicationManager.getApplication().messageBus
      .connect(this).subscribe(ProjectManager.TOPIC, this)
}
```

yonip23 (approved):
lgtm, looks like the adaptation was pretty smooth 🚀 
had some minor comments

dimacodota (approved):
lgtm

PR #567, opened 2023-07-09
DEV2-2972 add open devtools action

yonip23:
Since we need it to debug user complaints, and its already enabled by default in vscode, there's no harm in allowing this in jb as well.

https://github.com/codota/tabnine-intellij/assets/67855609/3e0d64a3-5b7d-4d01-8b7a-1fd783bc7052


dimacodota (approved):
love it ❤️

PR #569, opened 2023-07-10
DEV2-2972 remove reload action

yonip23:
it doesn't help with broken UI issues, so there's no reason to keep it


dimacodota (approved):
lgtm

PR #575, opened 2023-07-13
DEV2-3148: handle chat proxy

yairco1990:


Common/src/main/java/com/tabnineCommon/chat/ChatBrowser.kt
```diff
@@ -161,7 +156,28 @@ class ChatBrowser(messagesRouter: ChatMessagesRouter, private val project: Proje
...
+fun getProxySettings(): String? {
+    val httpConfigurable = HttpConfigurable.getInstance()
+
+    return if (httpConfigurable.USE_HTTP_PROXY) {
+        val auth = if (httpConfigurable.PROXY_AUTHENTICATION && httpConfigurable.proxyLogin?.isNotEmpty() == true) {
+            "${httpConfigurable.proxyLogin}:${httpConfigurable.plainProxyPassword}@"
+        } else {
+            ""
+        }
+        "${auth}${httpConfigurable.PROXY_HOST}:${httpConfigurable.PROXY_PORT}"
```
dimacodota, line 179: I think this is missing the protocol(http/https)

dimacodota (commented):
See my comment

Common/src/main/java/com/tabnineCommon/chat/ChatBrowser.kt
yairco1990: but there is no option in IJ settings to choose a protocol. by omitting the protocol, it should use http.

dimacodota: Yes, we use http for the binary

dimacodota: https://github.com/codota/tabnine-intellij/blob/1addb608344be27f228518a883420a22046f3e53/Common/src/main/java/com/tabnineCommon/binary/BinaryProcessGateway.kt#L56

yairco1990: but what I did is part of the chromium protocol:
https://www.chromium.org/developers/design-documents/network-settings/#command-line-options-for-proxy-settings
we can set "http=...", but "http://..." doesn't seem to be the protocol in chromium. (http is the default so we don't have to specify it)

dimacodota (approved):

PR #577, opened 2023-07-16
DEV2-3185 use only visible range in editor context

yonip23:


Common/src/main/java/com/tabnineCommon/chat/commandHandlers/GetEditorContextHandler.kt
```diff
@@ -37,20 +38,10 @@ class GetEditorContextHandler(gson: Gson) : ChatMessageHandler<Unit, GetEditorCo
     private val binaryRequestFacade = DependencyContainer.instanceOfBinaryRequestFacade()
 
     override fun handle(payload: Unit?, project: Project): GetEditorContextResponsePayload {
-        val editor = getEditorFromProject(project)
+        val editor = getEditorFromProject(project) ?: return noEditorResponse(project)
```
yonip23, line 41: just extracted the code inside the `if (editor == null)` block

dimacodota (approved):
lgtm

PR #598, opened 2023-07-30
DEV2-3197 breakdown editor context to dynamic context parts

yonip23:


yonip23:
TODO: bump chat version once its released

yairco1990 (approved):
lgtm

Common/src/main/java/com/tabnineCommon/chat/commandHandlers/context/DiagnosticsContext.kt
```diff
@@ -0,0 +1,54 @@
...
+) : EnrichingContextData {
+    // Used for serialization - do not remove
+    private val type: EnrichingContextType = EnrichingContextType.Diagnostics
+
+    companion object {
+        fun create(editor: Editor, project: Project): DiagnosticsContext? {
+            val visibleRange = getVisibleRange(editor) ?: return null
+            val highlights = mutableListOf<String>()
+
+            DaemonCodeAnalyzerImpl.processHighlights(
```
dimacodota, line 23: can you check if this can throw an exception ?

yonip23: it doesn't say so in the method declaration, but probably can throw an exception if you don't have read access or something.
anyway, if it throws an exception, we will send an error message to the chat and the chat knows how to handle it.
good enough IMO, wdyt?

dimacodota (approved):
lgtm

PR #607, opened 2023-08-07
DEV2-3336 add intent to conversation state

yonip23:
since the latest fixes to the intent detection mechanism, the UI relies on the `intent` field of the conversation state to render the chat history (combines the intent to the user's question to create the entire title of the chat item)
for example, here it joins "/explain-code" to "with text" for the title: 
image

since kotlin&java are strict with serialization, we need to add this field explicitly in order for it to be persisted

### DIFF
--- a/Common/src/main/java/com/tabnineCommon/chat/commandHandlers/chatState/ChatState.kt
+++ b/Common/src/main/java/com/tabnineCommon/chat/commandHandlers/chatState/ChatState.kt
@@ -5,7 +5,15 @@ import com.intellij.ide.util.PropertiesComponent
 
 private const val CHAT_CONVERSATIONS_KEY = "com.tabnine.chat.conversations.v2"
 
-data class ChatMessageProps(val id: String, val text: Any, val isBot: Boolean, val timestamp: String, val selected: Int?)
+data class ChatMessageProps(
+    val id: String,
+    val text: Any,
+    val isBot: Boolean,
+    val timestamp: String,
+    val selected: Int?,
+    val intent: Any?
+)
+
 data class ChatConversation(val id: String, val messages: List<ChatMessageProps>)
 
 data class ChatStateData(val conversations: MutableMap<String, ChatConversation>)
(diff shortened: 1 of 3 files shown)


Common/src/main/java/com/tabnineCommon/chat/commandHandlers/chatState/ChatState.kt
```diff
@@ -21,7 +29,8 @@ class ChatState(private val gson: Gson) {
     }
 
     fun get(): ChatStateData {
-        return PropertiesComponent.getInstance().getValue(CHAT_CONVERSATIONS_KEY)?.let { gson.fromJson(it, ChatStateData::class.java) } ?: ChatStateData(mutableMapOf())
+        return PropertiesComponent.getInstance().getValue(CHAT_CONVERSATIONS_KEY)
+            ?.let { gson.fromJson(it, ChatStateData::class.java) } ?: ChatStateData(mutableMapOf())
```
yonip23, line 33: formatting

Common/src/main/java/com/tabnineCommon/chat/commandHandlers/chatState/ChatState.kt
```diff
@@ -5,7 +5,15 @@ import com.intellij.ide.util.PropertiesComponent
...
 private const val CHAT_CONVERSATIONS_KEY = "com.tabnine.chat.conversations.v2"
 
-data class ChatMessageProps(val id: String, val text: Any, val isBot: Boolean, val timestamp: String, val selected: Int?)
+data class ChatMessageProps(
+    val id: String,
+    val text: Any,
+    val isBot: Boolean,
+    val timestamp: String,
+    val selected: Int?,
+    val intent: Any?
```
dimacodota, line 14: this is backward compatible, right?

yonip23: i added serialization tests that verify precisely these cases of backwards compatibility in serialization - we're working with the same serializer in the entire plugin.
see my [next commit here](https://github.com/codota/tabnine-intellij/pull/607/commits/3bd98426b962c0665de8b626605c736a0d9d124c)

dimacodota (approved):
LGTM

PR #609, opened 2023-08-08
DEV2-3340 include text in workspace context

yonip23:
1. fetch at most 2 symbols per "findSymbols" command
2. instead of only `"name - file path"`, add the entire symbol code, like the retrieval does
3. cache the `BasicContext` to later be used by the "get_enriching_context" handler
4. try to obtain the language field from the binary's metadata response, and use the IDE one as fallback (for both basic and enriching context handlers)

### DIFF
--- a/Common/src/main/java/com/tabnineCommon/chat/commandHandlers/context/GetBasicContextHandler.kt
+++ b/Common/src/main/java/com/tabnineCommon/chat/commandHandlers/context/GetBasicContextHandler.kt
@@ -11,9 +11,9 @@ import com.tabnineCommon.general.DependencyContainer
 import java.io.File
 
 data class BasicContext(
-    private val fileUri: String? = null,
-    private val language: String? = null,
-    private var metadata: JsonObject? = null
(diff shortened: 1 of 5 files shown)


Common/src/main/java/com/tabnineCommon/chat/commandHandlers/context/GetBasicContextHandler.kt
```diff
@@ -23,20 +23,23 @@ data class BasicContext(
...
 
         val psiFile = PsiDocumentManager.getInstance(project).getPsiFile(editor.document)
         val fileUri = psiFile?.virtualFile?.path
-        val language = psiFile?.language?.id
-
-        var metadata = if (fileUri != null) binaryRequestFacade.executeRequest(FileMetadataRequest(fileUri)) else null
-
+        var metadata = fileUri?.let {
+            binaryRequestFacade.executeRequest(FileMetadataRequest(it))
+        }
```
yonip23, line 33: same logic - used `?.let` instead of `if/else`

PR #611, opened 2023-08-09
DEV2-3340 search in both snake case and camel case

yonip23:


Common/src/main/java/com/tabnineCommon/chat/commandHandlers/context/workspace/FindSymbolsCommandExecutor.kt
```diff
@@ -2,11 +2,28 @@ package com.tabnineCommon.chat.commandHandlers.context.workspace
...
+        val camelCaseSymbols = SymbolsResolver.resolveSymbols(
+            project, editor.document, camelCaseArg,
+            MAX_RESULTS_PER_SYMBOL
+        )
+        val snakeCaseSymbols = SymbolsResolver.resolveSymbols(
+            project, editor.document, snakeCaseArg,
+            MAX_RESULTS_PER_SYMBOL
+        )
+
+        return Stream.concat(camelCaseSymbols.stream(), snakeCaseSymbols.stream())
```
dimacodota, line 25: i think you should  dedup that list

Common/src/main/java/com/tabnineCommon/chat/commandHandlers/context/workspace/FindSymbolsCommandExecutor.kt
```diff
@@ -2,11 +2,28 @@ package com.tabnineCommon.chat.commandHandlers.context.workspace
...
 class FindSymbolsCommandExecutor : CommandsExecutor {
     override fun execute(arg: String, editor: Editor, project: Project): List<String> {
-        return SymbolsResolver.resolveSymbols(project, editor.document, arg, 5)
+        val camelCaseArg = StringCaseConverter.toCamelCase(arg)
+        val snakeCaseArg = StringCaseConverter.toSnakeCase(arg)
+        val camelCaseSymbols = SymbolsResolver.resolveSymbols(
+            project, editor.document, camelCaseArg,
+            MAX_RESULTS_PER_SYMBOL
+        )
+        val snakeCaseSymbols = SymbolsResolver.resolveSymbols(
```
dimacodota, line 20: is this case sensitive?

Common/src/main/java/com/tabnineCommon/chat/commandHandlers/context/workspace/FindSymbolsCommandExecutor.kt
yonip23: i added a dedup at the end in `WorkspaceContext.kt`: `WorkspaceContext(symbols.distinct())`
its better to do it there, because duplicate symbols can arrive from different commands, for example if one command is "foo" and the other is "fooBar" - you can find "fooBar" in both.

Common/src/main/java/com/tabnineCommon/chat/commandHandlers/context/workspace/FindSymbolsCommandExecutor.kt
yonip23: yes:
image
image

PR #612, opened 2023-08-10
DEV2-3340 run both naming conventions in parallel

yonip23:
following up on [this comment](https://github.com/codota/tabnine-intellij/pull/611#discussion_r1289133119) from #611,
i think its ok to use `submitReadAction` in both locations, just like nested `Promise.all` are ok:
The code in `submitReadAction` is:
```kotlin
fun <T> submitReadAction(action: () -> T): CompletableFuture<T> {
    val future = CompletableFuture<T>()
    AppExecutorUtil.getAppExecutorService().submit {
        ApplicationManager.getApplication().runReadAction {
            try {
                val result = action()
                future.complete(result)
            } catch (e: Throwable) {
                future.completeExceptionally(e)
            }
        }
    }
    return future
}
```

- `AppExecutorUtil.getAppExecutorService().submit` -> this creates a task to be executed by the executor service, which is practically a thread pool
- `ApplicationManager.getApplication().runReadAction` -> according to their documentation:
> Runs the specified read action. Can be called from any thread. The action is executed immediately if no write action is currently running, or blocked until the currently running write action completes.
So it seems that this read action is not creating another thread, but rather behaves like Rust's `RwLock`.

Overall, it looks like the `submitReadAction` API is requiring a single thread from the threadpool, and a single read-write lock. So I think its ok to use it in nested futures.

### DIFF
--- a/Common/src/main/java/com/tabnineCommon/chat/commandHandlers/context/workspace/FindSymbolsCommandExecutor.kt
+++ b/Common/src/main/java/com/tabnineCommon/chat/commandHandlers/context/workspace/FindSymbolsCommandExecutor.kt
@@ -14,17 +14,20 @@ class FindSymbolsCommandExecutor : CommandsExecutor {
         val camelCaseArg = StringCaseConverter.toCamelCase(arg)
         val snakeCaseArg = StringCaseConverter.toSnakeCase(arg)
 
-        val tasks = listOf(submitReadAction {
-            SymbolsResolver.resolveSymbols(
-                project, editor.document, camelCaseArg,
-                MAX_RESULTS_PER_SYMBOL
-            )
-        }, submitReadAction {
-            SymbolsResolver.resolveSymbols(
-                project, editor.document, snakeCaseArg,
-                MAX_RESULTS_PER_SYMBOL
-            )
-        })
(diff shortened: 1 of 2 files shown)


Common/src/main/java/com/tabnineCommon/chat/commandHandlers/context/workspace/FindSymbolsCommandExecutor.kt
```diff
@@ -4,25 +4,34 @@ import com.intellij.openapi.editor.Editor
...
+            submitReadAction {
+                SymbolsResolver.resolveSymbols(
+                    project, editor.document, snakeCaseArg,
+                    MAX_RESULTS_PER_SYMBOL
+                )
+            }
         )
 
-        return Stream.concat(camelCaseSymbols.stream(), snakeCaseSymbols.stream())
+        CompletableFuture.allOf(*tasks.toTypedArray()).get()
```
yonip23, line 32: im not putting a TO on this awaiting, since the upper layer already caps this operation with a TO (2.5s).

yairco1990 (approved):
lgtm

dimacodota (approved):
lgtm

PR #619, opened 2023-08-16
DEV2-3471 fix read/write permissions of chat message handlers

yonip23:
This PR changes the asynchronous thread dispatching and read/write permissions acquired by the chat handlers.
The motivation behind this change is - the permissions required to run each handler are different, and each handler should decide on its permissions.

### Before
- All handlers were running with write actions on the AWT thread (by `ChatBrowser::handleIncomingMessage`)
- `GetEnrichingContextHandler` was dispatching its specific context creators (editor/workspace/diagnostics) with read actions on a thread from JB's thread pool (`AppExecutorUtil.getAppExecutorService().submit`)
  The context creators can in turn spawn nested async tasks on the same thread pool, with read permissions as well.

### After
- The handlers are being executed on the same thread as the `ChatBrowser`, no async dispatching there.
- Each handler decides whether it needs to spawn async tasks and which permissions it needs.
- `GetBasicContextHandler`, which didn't handle any permissions in master, now runs with read lock (still sync)
- `GetEnrichingContextHandler` is not deciding on the asynchronicity and permissions of its context creators.
- The context creators are now running async-ly, each one decides whether it needs read/write permissions. (Diagnostics requires write permissions, the rest require read permissions).

To provide a convenient mechanism for running an async task with the correct permissions, I created the `AsyncAction` class, which accepts the required permission (`READ`/`WRITE`) - and executes the provided `action` accordingly (see details in the class' documentation).

### DIFF
--- a/Common/src/main/java/com/tabnineCommon/chat/ChatBrowser.kt
+++ b/Common/src/main/java/com/tabnineCommon/chat/ChatBrowser.kt
@@ -1,6 +1,5 @@
 package com.tabnineCommon.chat
 
-import com.intellij.openapi.application.ApplicationManager
 import com.intellij.openapi.diagnostic.Logger
 import com.intellij.openapi.project.Project
 import com.intellij.ui.jcef.JBCefApp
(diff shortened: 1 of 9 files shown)


ofekby (approved):
nice, lgtm

dimacodota (approved):
lgtm 🚀

PR #620, opened 2023-08-16
DEV2-3375 - update Enterprise Plugin name and icon to avoid confusion

avichay77:
DEV2-3375
1. Rename enterprise plugin so it will be clear it's for Enterprise Self Hoster
2. Replace icon for Enterprise plugin
3. Add disclaimer that this is not for Enterprise SaaS
4. Update permissive licenses list
5. remove discord/slack links


yonip23 (commented):
the `TabnineSelfHosted` icon should not be the download icon (with the down arrow), only the marketplace plugin should have it

dimacodota (approved):

PR #634, opened 2023-09-04
DEV2-3610: Pass server URL only if self hosted

assapir:


Common/src/main/java/com/tabnineCommon/chat/commandHandlers/InitHandler.kt
```diff
@@ -24,8 +25,12 @@ class InitHandler(gson: Gson) : ChatMessageHandler<Unit, InitPayload>(gson) {
         val colorPalette = readColorPalette()
         val isDarkTheme = EditorColorsManager.getInstance().isDarkEditor
         val font = EditorColorsManager.getInstance().globalScheme.getFont(EditorFontType.PLAIN).size
-
-        return InitPayload("ij", isDarkTheme, colorPalette, font, isTelemetryEnabled(), getServerUrl())
+        val serverUrl = if (Config.IS_SELF_HOSTED) {
+            getServerUrl()
```
yonip23, line 29: Why isn't this condition baked into the `getServerUrl` method?

assapir: no, but I get your comment. will move it

yonip23 (approved):

yanir-codota (approved):

PR #643, opened 2023-09-14
DEV2-3719: Fix URL when downloading self hosted version

assapir:
There were serval bugs:
1. The updater didn't set the URL after the user set it.
2. Which made it not to add the custom repository
3. It was permanently replacing the same value

I think this needs to be published to the marketplace and fix the version in the 1.2.0 branch, is the code also called from there

### DIFF
--- a/Common/src/main/java/com/tabnineCommon/general/Utils.java
+++ b/Common/src/main/java/com/tabnineCommon/general/Utils.java
@@ -131,7 +131,7 @@ public final class Utils {
   }
 
   public static String trimEndSlashAndWhitespace(String text) {
-    return text.replace("/\\s*$", "");
+    return text.replace("/*\\s*$", "");
   }
 
   public static void setCustomRepository(String url) {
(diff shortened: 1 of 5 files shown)


yanir-codota (approved):
LGTM

Common/src/main/java/com/tabnineCommon/general/Utils.java
```diff
@@ -131,7 +131,7 @@ public static boolean isUnitTestMode() {
   }
 
   public static String trimEndSlashAndWhitespace(String text) {
-    return text.replace("/\\s*$", "");
+    return text.replace("/*\\s*$", "");
```
yonip23, line 134: what does it fix?

yonip23 (approved):
lgtm, just a small question

Common/src/main/java/com/tabnineCommon/general/Utils.java
assapir: multiple `///` in the end

PR #644, opened 2023-09-14
DEV2-3719: Apply the fix also for regular self hosted

assapir:
The problem is that the `AppSettingsState` is different type than the one in for the marketplace, so for that to work it have to come as a callback. also applied the fixes to the save function


yonip23 (approved):

yanir-codota (approved):
LGTM

PR #652, opened 2023-10-08
DEV2-3931 add quick fix action

yonip23:
Add a "Fix using Tabnine" quick-action
Screenshot from 2023-10-08 15-50-31

### Details
1. This action opens the chat window and sends it the text `"/fix-code"`.
2. This action only appears if there's at least one warning/error level diagnostic in the selected code range, or - if no code is selected - on the PSI element the cursor is on.


Common/src/main/java/com/tabnineCommon/chat/actions/common/ChatActionCommunicator.kt
```diff
@@ -0,0 +1,54 @@
+package com.tabnineCommon.chat.actions.common
```
yonip23, line 1: extracted logic from `AskChatAction`, to be used in `TabnineQuickFixAction` (and potentially future actions) as well.

dimacodota (approved):
lgtm

PR #653, opened 2023-10-08
DEV2-3931 fix chat first loading

yonip23:
### Problem
An action can't invoke the chat tool window before the user manually opens it in the current session.

It happens because `ChatBrowser.getInstance` would return `null` until the user opens the tool window, thus creating a browser instance.

### Solution
I solved it by making the browser creation available from anywhere, not just from the user's manual hook.
This solution required the following:
1. Making the `ChatMessagesRouter` singleton and not pass it in the creation phase.
2. Changing `ChatBrowser.getInstance` to create an instance if it doesn't exist on the project's user data, and making the constructor private to prevent double-creation per project.
3. Changing the "browser loaded" listeners to "chat app startup" listeners
   This was necessary because the previous implementation simply didn't work: The browser is considered "loaded" before the app itself is running, just when the static files are loaded onto the browser instance.
   I solved it by hooking the mechanism to the first message received from the app instead of the low-level-ish browser loading API, which means that the react app is completely running.
   And since we're sending requests from the app as soon as it's alive, it's a pretty good approximation.

### DIFF
--- a/Common/src/main/java/com/tabnineCommon/chat/ChatBrowser.kt
+++ b/Common/src/main/java/com/tabnineCommon/chat/ChatBrowser.kt
@@ -18,32 +18,35 @@ import java.awt.datatransfer.StringSelection
 import java.nio.file.Paths
 import java.util.concurrent.atomic.AtomicBoolean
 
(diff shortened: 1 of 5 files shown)


dimacodota (approved):
lgtm

PR #666, opened 2023-10-25
DEV2-4026: run chat's init handler inside an EDT

yairco1990:


Common/src/main/java/com/tabnineCommon/chat/commandHandlers/InitHandler.kt
```diff
@@ -21,11 +22,16 @@ data class InitPayload(
 
 class InitHandler(gson: Gson) : ChatMessageHandler<Unit, InitPayload>(gson) {
     override fun handle(payload: Unit?, project: Project): InitPayload {
-        val colorPalette = readColorPalette()
-        val isDarkTheme = EditorColorsManager.getInstance().isDarkEditor
-        val font = EditorColorsManager.getInstance().globalScheme.getFont(EditorFontType.PLAIN).size
+        var result: InitPayload? = null
```
dimacodota, line 25: why not return a future hare and manage the flow async?

or at least i think there should be thread safe  wrapper around this

dimacodota (approved):
lgtm

PR #671, opened 2023-10-31
DEV2-4020 login with custom token

amirbilu:


Common/src/main/java/com/tabnineCommon/binary/requests/login/LoginWithCustomTokenResponse.kt
```diff
@@ -0,0 +1,9 @@
+package com.tabnineCommon.binary.requests.login
+
+import com.google.gson.annotations.SerializedName
+import com.tabnineCommon.binary.BinaryResponse
+
+data class LoginWithCustomTokenResponse(
+    @SerializedName("is_success")
+    var isSuccess: Boolean? = null,
```
yanir-codota, line 8: if this fails, will we have any ability to debug it?

yanir-codota (approved):
LGTM. 
Had a question regarding error resolving/debug

Common/src/main/java/com/tabnineCommon/binary/requests/login/LoginWithCustomTokenResponse.kt
amirbilu: currently no.. let's talk about this tomorrow

PR #678, opened 2023-11-05
DEV2-4083: Update wiremock dependency to resolve security issue

ofekby:


yanir-codota (commented):
I don't know what this does. 
can you elaborate?

ofekby:
> I don't know what this does. can you elaborate?

It had a security issue, updated the library

PR #680, opened 2023-11-08
get editor context with read permission

yairco1990:


yonip23 (approved):

Common/src/main/java/com/tabnineCommon/chat/commandHandlers/ChatMessageHandler.kt
```diff
@@ -18,7 +18,7 @@ abstract class ChatMessageHandler<RequestPayload, ResponsePayload>(protected val
     }
 
     protected fun getEditorFromProject(project: Project): Editor? {
-        return AsyncAction(ActionPermissions.WRITE).execute {
+        return AsyncAction(ActionPermissions.READ).execute {
```
dimacodota, line 21: @yairco1990 why not reuse the `getActiveEditor` method that is used in the completions? 
https://github.com/codota/tabnine-intellij/blob/ee546ce2ea61b06e3f0662b965896ea3ae73ba38/Common/src/main/java/com/tabnineCommon/inline/TabnineDocumentListener.java#L95

yairco1990: the completions approach also uses EDT which can lead to the same problem

dimacodota (approved):
lgtm

PR #681, opened 2023-11-08
DEV2-4002: calculate workspace language

yonidavidson:
Same code as vscode, by using the hard-coded list of languages we have better chance to find the language also in IDE's which are not fully configured.

### DIFF
--- a/Common/src/main/java/com/tabnineCommon/chat/commandHandlers/context/GetBasicContextHandler.kt
+++ b/Common/src/main/java/com/tabnineCommon/chat/commandHandlers/context/GetBasicContextHandler.kt
@@ -5,7 +5,10 @@ import com.google.gson.JsonElement
 import com.google.gson.JsonObject
 import com.intellij.openapi.application.ReadAction
 import com.intellij.openapi.project.Project
+import com.intellij.openapi.project.ProjectManager
 import com.intellij.psi.PsiDocumentManager
+import com.intellij.psi.PsiManager
+import com.intellij.util.indexing.FileBasedIndex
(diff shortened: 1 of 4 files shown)


Common/src/main/java/com/tabnineCommon/chat/commandHandlers/context/GetBasicContextHandler.kt
```diff
@@ -54,7 +56,113 @@ class GetBasicContextHandler(gson: Gson) : ChatMessageHandler<Unit, BasicContext
...
-        return BasicContext(metadata)
+        return BasicContext(null, getPredominantWorkspaceLanguage(), metadata)
     }
 }
+
+fun getPredominantWorkspaceLanguage(
+    includeFilePredicate: (String) -> Boolean = { true }
+): String? {
+    val project = ProjectManager.getInstance().openProjects.firstOrNull() ?: return null
+    val fileIndex = FileBasedIndex.getInstance()
```
dimacodota, line 67: pls, check that it is running in the correct context (read/write) https://plugins.jetbrains.com/docs/intellij/general-threading-rules.html#read-access

dimacodota (approved):
lgtm, see  comment

Common/src/main/java/com/tabnineCommon/chat/commandHandlers/context/GetBasicContextHandler.kt
yonidavidson: I can make it run in read 
```kotlin
ApplicationManager.getApplication().runReadAction {
```

PR #685, opened 2023-11-12
DEV2-4218 add support for multi-repo workspace

yonip23:
I ran on IJ CE 2020.2 (the sandbox, by running `runIde`) and on IJ CE 2023.5 (local installation) to make sure there were no EDT/read violations.

one concern i have, is that the `"Workspace"` API returns null, and we don't count null responses as success.. maybe I should return "Done" from the binary or something?


Common/src/main/java/com/tabnineCommon/chat/commandHandlers/InitHandler.kt
```diff
@@ -15,18 +18,24 @@ data class InitPayload(
...
+
         return InitPayload(
             "ij",
             EditorColorsManager.getInstance().isDarkEditor,
             readColorPalette(),
             EditorColorsManager.getInstance().globalScheme.getFont(EditorFontType.PLAIN).size,
             isTelemetryEnabled(),
-            getServerUrl()
+            getServerUrl(),
+            workspaceFolders
```
amircodota, line 38: why are we send the workspace folders with the init?

amircodota: I mean, this list could change after the chat is open.

Wouldn't it be better to sk for it when you issue a retrieval query?

amircodota (commented):
see my comments

Common/src/main/java/com/tabnineCommon/chat/commandHandlers/InitHandler.kt
yonip23: huh, it was just some draft i had when i played around with the workspace folders last week, i didn't notice i hadn't rolled it back.

amircodota (approved):
LGTM

yonip23:
waiting for the binary to be released before merging this

amircodota (approved):
LGTM

PR #689, opened 2023-11-19
DEV2-4267 Fix chat screen

amirbilu:


Common/src/main/java/com/tabnineCommon/chat/ChatFrame.kt
```diff
@@ -32,6 +34,7 @@ class ChatFrame(private val project: Project, private val binaryRequestFacade: B
 
     init {
         layout = BorderLayout()
+        isLoggedIn = ServiceManager.getService(BinaryStateService::class.java).lastStateResponse?.isLoggedIn ?: false
```
yonip23, line 37: what happens if the `isLoggedIn` value changes after this line is executed? how do we know to re-render the chat?

yonip23 (approved):

PR #691, opened 2023-11-20
DEV2-3893 filter unwanted workspace root paths

yonip23:


Common/src/main/java/com/tabnineCommon/lifecycle/WorkspaceListenerService.kt
```diff
@@ -57,6 +58,14 @@ class WorkspaceListenerService {
                 Logger.getInstance(javaClass).debug("$url in project ${project.name} has unsupported protocol (${url.protocol})")
                 continue
             }
+            if (rootPaths.any { it.startsWith(url.path) }) {
```
amircodota, line 61: If the rootPath appears before its parent it would still be included. Better check this after you have the full list

amircodota (commented):
see my comment

yonip23:
dedup test: 
image

amircodota (approved):
LGTM

PR #699, opened 2023-12-04
DEV2-4411: code lens

yairco1990:
https://github.com/codota/tabnine-intellij/assets/10290661/4cf54d21-5caa-4579-ac06-eb57d831c10a


yairco1990:
@amircodota testing this feature is not easy. I spent some time writing tests, but it will take time to finish them.
regarding the `jvm-default=enabled` flag, is because of this error:
`Inheritance from an interface with '@JvmDefault' members is only allowed with -Xjvm-default option`
it refers to inheritance from ImmediateConfigurable which has @JvmDefault annotations.

amircodota (approved):
LGTM

